PR #25, opened 2025-07-28
Unified configuration RFC

YurBoiRene:


docs/source/rfcs/3-unified-configuration.rst
```diff
@@ -0,0 +1,409 @@
...
+        /// Must match the ID given from [`ProcessorImplSpec`]
+        id: String
+        /// Optional arbitrary configuration
+        config: Option<serde_yaml::Value>,
+    }
+
+
+.. _ProcessorBuilderConfigMapper:
+
+ProcessorBuilderConfigMapper (Name TBD)
```
lockbox: for this config mapper, I'm assuming here that you'd collect using `inventory` or something to build these `Vec<..Spec>`'s?

lockbox: got confused from the example usage up top and this description, yup I found it

docs/source/rfcs/3-unified-configuration.rst
```diff
@@ -0,0 +1,409 @@
...
+PCS subscribes to the ``from`` processor's ``BytesMessage`` stream and will
+route Tx messages on the ``from`` processor port into ``Receive`` calls in the
+``to`` processor's port.
+
+Direction is referred to from the ``from`` processor's perspective. I.e.
+``direction: tx`` would mean the ``from(tx) -> to(rx)`` connection is made by
+the reverse connection is not made.
+
+SPI
+"""
```
lockbox, line 319: Yeah, thoughts for `SPI` based on #28, eventually maybe we could do something like this where the configuration itself actually builds an emulator, for this problem for now this is too verbose, as it has the actual types of the peripherals etc in here, but I wanted to keep this not around somewhere since it could be pretty useful at a later point. 

```yaml
- name: example proc
  processor:
    ...
    peripherals:
    - name: spi0
      type: PPC405Spi
      port: c
      configuration:
      - mode: slave
      - master-node: example-master-processor
      - master-port: 8

```

I think that the proposed spi configuration in your example in the next section solves the current problem and is what we should do 👍

docs/source/rfcs/3-unified-configuration.rst
```diff
@@ -0,0 +1,409 @@
...
+        port: 1
+      slaves:
+        - processor: processor_2
+          # refers to chip select id of master, usually gpio number
+          chip_select: 0
+          port: 1
+    - !I2C
+      master: 
+        processor: processor_1
+        address: 0x14
```
lockbox: I don't think we should be specifying addresses in here as a mandatory thing. the processor or device implementation should probably be the one to set this at runtime. We could maybe add an `address_hint` or something that is a default? IDK

docs/source/rfcs/3-unified-configuration.rst
```diff
@@ -0,0 +1,409 @@
...
+      name: Processor 2
+      host: localhost
+      port: 18001
+  connections:
+    - !Uart
+      from: 
+        processor: processor_1
+        # name override
+        name: Processor 1 Uart0
+        port: uart0
```
lockbox: instead of "uart0", port names / numbers are a string that looks like "0", "1", "A", "B" etc (on some platforms its "port A", on some its "port 1")

docs/source/rfcs/3-unified-configuration.rst
```diff
@@ -0,0 +1,409 @@
...
+        - processor: processor_2
+          # refers to chip select id of master, usually gpio number
+          chip_select: 0
+          port: 1
+    - !I2C
+      master: 
+        processor: processor_1
+        address: 0x14
+        port: 1
+      slaves:
```
lockbox, line 459: for the i2c slaves it would be nice to also allow for some of the devices we have to be able to be specified and referenced here like

```yaml
slaves:
  - device: cool device
    type: RTC
    device_type: Rtc1022A
    configuration:
    - arg1: val1
    - arg2: val2
```

docs/source/rfcs/3-unified-configuration.rst
```diff
@@ -0,0 +1,409 @@
...
+TODO: I feel like the processor should define the i2c device address somehow,
+not the peripheral communication config.
+
+
+Drawbacks/Alternatives
+======================
+
+
+Future Work
+===========
```
lockbox, line 471: This configuration stuff will be a nice forcing function for the metadata / configuration stuff we've bee waffling over forever 😆 I'm excited

docs/source/rfcs/3-unified-configuration.rst
```diff
@@ -0,0 +1,409 @@
...
+      master: 
+        processor: processor_1
+        address: 0x14
+        port: 1
+      slaves:
+        - processor: processor_2
+          address: 0x15
+          port: 1
+
+TODO: I feel like the processor should define the i2c device address somehow,
```
lockbox, line 463: Something not explicitly mentioned in this RFC (correct me if I'm wrong) that I think we should include well documented in the implementation for this:

the ability to map things in *after* parsing / loading / mapping this configuration yaml. meaning that you could roughly do:

- setup the processor mapper you want
- load / register the configuration
- programmatically add to the configuration (modify at a later date?)
- finalize / apply the configuration

This would allow us to treat emulators kind of like programs, where you have your solid / robust stuff baked into your code, you can load a configuration file, and then override things with command line options or environment variables.

This allows easy manipulation by bindings or other use code that is a more dynamic / optional or opt-in behavior and provides a lot more flexibility for not that much more work (imo).

docs/source/rfcs/3-unified-configuration.rst
```diff
@@ -0,0 +1,409 @@
...
+are passed as an ID for the processor implementation. More on this in
+:ref:`custom-implementations`.
+
+Second, a :ref:`ProcessorBuilderConfigMapper` is constructed and additional
+:ref:`ProcessorImplSpec`, ExecutorSpec, and PluginSpec can be registered here.
+
+Finally, the :ref:`ProcessorBuilderConfig` is combined with the
+:ref:`ProcessorBuilderConfigMapper` to create a ProcessorBuilder or Processor
+(TODO does this matter?).
+
```
lockbox, line 61: I think the naming should be adjusted here, since we can contract arbitrary processors from this. Using the terminology from the core-concepts docs this should probably be either a `Machine...` or we should call this like a `Workspace...`. I'm leaning towards the latter, since it leaves a little more wiggle room to give us expansion behind the scenes to add more options without the name being as confusing.

Naming the top level stuff `Workspace...` also gives us more freedom of movement with less breaking changes as we continue to flesh out what configurations, machines, external program integration etc. looks like from this POV. (eg. adding a "fuzzing profile/config" to a workspace wouldn't require the user importing/using new types or anything, and remains the top-level, regardless what the user is attempting to do since this already has it configuring an array of items, the `Processor` stuff being the top level, so even the yaml config might not necessarily have breaking changes as the project continues to evolve)

docs/source/rfcs/3-unified-configuration.rst
```diff
@@ -0,0 +1,409 @@
...
+Loading Programs
+^^^^^^^^^^^^^^^^
+
+We will use the ParametrizedLoader to load programs.
+
+One concern is that the ParametrizedLoader uses file paths to specify programs
+to load, if the yaml config is delivered remotely in the future this wouldn't
+make sense.
+
+There could be some exploratory research in defining and registering loadable
```
lockbox: I agree this should be tossed in as "future work". probably a precursor to the remote processors

docs/source/rfcs/3-unified-configuration.rst
```diff
@@ -0,0 +1,409 @@
...
+.. _Inventory: https://docs.rs/inventory/latest/inventory/
+
+Upsides:
+
+1. Super easy to register specs
+
+Downsides:
+
+1. Dependency on Inventory crate
+2. Inventory crate works on most OSes and WASM but does some hacky linker
```
lockbox: IMO its not that tacky, ctors are ubiquitous and would also work cross-language boundaries, whereas something like <https://github.com/dtolnay/linkme> is properly hacky and would probably cause a lot of pain. I think this is a good approach / balance of things to get the job done, and something we would have ended up with anyways after some time.

docs/source/rfcs/3-unified-configuration.rst
```diff
@@ -0,0 +1,409 @@
...
+Configuration of plugins, backends, executors, and processor impls is
+important. The current Styx bindings lack configuration just because it is hard
+to implement in a language agnostic way without taking lots of developer time.
+
+For our Unified Configuration, we propose a `config` item next to the `id` that
+can be optionally supplied to provide arbitrary config data. Then, when
+registering with the config mapper, each arbitrary impl will be configured via
+this config structure.
+
+TODO: is this proper? or should configs be passed as inline yaml strings
```
lockbox: I think this is fine, and that the inline yaml would cause more pain than its worth. It would probably we worth it to document what the mapping from the id/config type you're submitting would get resolved to in yaml, or provide a helper macro / function you could call to emit the yaml representation to easier debug things.

docs/source/rfcs/3-unified-configuration.rst
YurBoiRene: This is valid. We will need to route packets based on address but I guess yeah slaves will have to tell the PCS what their address on subscribe.

docs/source/rfcs/3-unified-configuration.rst
YurBoiRene: What is the intention of the configuration here? Is the idea that the PCS would spin up that device instead of connecting to an already running processor? Or is the extra info just for documentation purposes

lockbox: more for the former reason, though it would also serve as nice documentation.

some examples that I had in mind when writing this look like:
- rtc, set the time scale or default parameters
- ADC, set an equation or mode for example data input feed
- "lora bridge" (over uart or something) give it arguments to connect to the network
- some other configurable thing (like to talk to another processor or external simulator etc) that would be a "device" to the processor etc that would could have an in-tree model for

docs/source/rfcs/3-unified-configuration.rst
lockbox: my thoughts exactly, yes the only thing that the transport layer needs to know here is that the devices need to be talking to each other, they can figure out addresses later (and sometimes have defaults?)

docs/source/rfcs/3-unified-configuration.rst
YurBoiRene: Sounds good. I like having both options too

docs/source/rfcs/3-unified-configuration.rst
YurBoiRene: I changed `ProcessorBuilderConfigMapper` to `WorkspaceConfigMap` which I think is cleaner and reflects its workspace level scope

docs/source/rfcs/3-unified-configuration.rst
YurBoiRene: Yes I had this in mind but we should mention it. I included a blurb for this:

> It's at this point that configurations are in a native Rust structure and can
be stored or modified by external sources. For example, a CLI tool may override
processor configuration options. 

Let me know if there is anything else this needs.

lockbox: that's good, anything more and it starts to look more like documentation for something that doesnt even exist yet :p

PR #26, opened 2025-07-28
Physical address hooks

YurBoiRene:


styx/bindings/styx-c-api/inc/styx_emulator.h
```diff
@@ -759,11 +759,8 @@ typedef struct StyxHook_InvalidInstructionData {
  * standard section memory permissions
  */
 typedef struct MemoryPermissions {
-  uint32_t bits;
+  Internal _0;
```
lockbox: i dont wanna know what crazy magic keeps changing this

styx/core/styx-cpu-pcode-backend/src/memory/space_manager.rs
```diff
@@ -324,17 +324,29 @@ impl SpaceManager {
             .info
             .endian;
         let mut data = cpu.get_value_mmu(mmu, varnode)?.to_bytes(endian);
+        let virtual_address = varnode.offset;
+        let physical_address = mmu.translate_va(
+            virtual_address,
+            MemoryOperation::Read,
+            MemoryType::Data,
```
lockbox, line 331: just a quick sanity check here: is this `MemoryType::Data` assumption going to remain valid here and in the `memory_write_hook`? My concern is that for things like `AVR` that have `code|data`, where memory could be *not* data, that this would then mess something up, since we now are hardcoding the mapping to strictly `MemoryType::Data`

styx/core/styx-processor/src/memory/tlb/closure.rs
```diff
@@ -14,15 +14,14 @@ use crate::memory::{
 /// perform a simple translation. Proper processor implementations should not use this and instead
 /// implement [TlbImpl].
 pub struct ClosureTlb {
-    pub closure: TranslateClosure,
+    pub closure: TranslateFn,
```
lockbox: why the rename? if the module is still called the original? (i agree w the change fwiw)

styx/bindings/styx-c-api/inc/styx_emulator.h
YurBoiRene: :angry:

styx/core/styx-cpu-pcode-backend/src/memory/space_manager.rs
YurBoiRene: You analysis is correct, however I don't see an alternative at this point? The pcode op load/store have no information on whether the operation is for code or data.

To support AVR I think there will be two spaces (`MmuDataSpace` and `MmuCodeSpace`) that are mapped to `SpaceName::Custom("DataRam")` and `SpaceName::Custom("CodeRam")` and the code space will do Mmu operations with `MemoryType::Code`.

styx/core/styx-processor/src/memory/tlb/closure.rs
YurBoiRene: I'll change the structure name/other docs too

PR #44, opened 2025-08-20
unified configuration implmentation

### LINKED ISSUE #82: Unified Configuration PoC
This should implement the Unified Configuration (uconf) section of RFC 3.

This impl can omit inventory support.

### LINKED ISSUE #83: Unified Configuration Inventory Support
Use [inventory](https://github.com/dtolnay/inventory) to add arbitrary impls for linked crates

YurBoiRene:
Closes #82 
Closes #83 

todo:
- [x] processors registered
- [x] plugins registered
- [x] documentation
- [x] multiple processors, one yaml

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5850,6 +5850,7 @@ dependencies = [
  "enum-map",
  "futures",
  "num_enum",
+ "serde",
  "styx-blackfin-sys",
  "styx-core",
  "styx-plugins",
@@ -6034,6 +6035,7 @@ dependencies = [
  "num-derive",
  "num-traits",
  "paste",
+ "serde",
  "styx-core",
  "styx-cyclone-v-hps-sys",
  "styx-gic",
@@ -6090,7 +6092,9 @@ name = "styx-debug-tools"
 version = "1.2.0"
 dependencies = [
  "async-trait",
+ "serde",
  "styx-core",
+ "styx-uconf",
  "styx-workspace-hack",
  "tokio",
  "tracing",
@@ -7794,6 +7798,7 @@ dependencies = [
  "opentelemetry-otlp",
  "opentelemetry_sdk",
  "styx-core",
+ "styx-uconf",
  "styx-workspace-hack",
  "tokio",
  "tracing",
--- a/styx/plugins/styx-debug-tools/Cargo.toml
+++ b/styx/plugins/styx-debug-tools/Cargo.toml
@@ -12,6 +12,8 @@ workspace = true
 
 [dependencies]
 styx-core = { workspace = true }
+serde = { workspace = true }
+styx-uconf = { path = "../../../incubation/styx-uconf" }
 
 async-trait = { workspace = true }
 tracing = { workspace = true }
--- a/styx/plugins/styx-debug-tools/src/lib.rs
+++ b/styx/plugins/styx-debug-tools/src/lib.rs
@@ -76,10 +76,12 @@ impl ProtectionFaultHook for HaltableHook {
 ///
 /// The plugin behavior is controllable, depending on the `halt` argument,
 /// the plugin will halt the target program execution when the hook is fired
-#[derive(Debug, Default)]
+#[derive(Debug, Default, serde::Deserialize)]
 pub struct UnmappedMemoryFaultPlugin {
     halt: bool,
 }
+
+styx_uconf::register_component_config!(register plugin: id = unmapped_memory_fault, component = UnmappedMemoryFaultPlugin);
 
 impl UnmappedMemoryFaultPlugin {
     pub fn new(halt: bool) -> Self {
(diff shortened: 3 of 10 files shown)


incubation/styx-uconf/src/components.rs
```diff
@@ -0,0 +1,184 @@
...
+    pub type executor = Box<dyn ExecutorImpl>;
+    pub type plugin = Box<dyn UninitPlugin>;
+}
+
+/// Compile time component for registration.
+///
+/// If you just want to register a component, see the module level documentation for
+/// [`crate::components`].
+///
+/// This is used integrally by the component registration macros and stores module/file information
```
lockbox: should `integrally` be `iternally`?

incubation/styx-uconf-cli/src/main.rs
```diff
@@ -0,0 +1,84 @@
...
+                        )
+                    })?;
+
+                let mut builder = styx_uconf::realize_unified(yaml)
+                    .with_context(|| "could not realize processor")?;
+                assert_eq!(
+                    builder.len(),
+                    1,
+                    "uconf cli only supports a single processor"
+                );
```
lockbox: this should be updated since we now support multiple processors

incubation/styx-uconf-cli/src/main.rs
```diff
@@ -0,0 +1,84 @@
...
+
+                let mut processor = builder
+                    .pop()
```
lockbox: nit: the `.pop()` here feels weird. could you instead use an iterator here and just foreach do xxxx? (now that there are multiple)

incubation/styx-uconf/src/components.rs
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        $crate::components::inventory_submit! {
+            $crate::$class!(stringify!($id), $id)
+        }
+    };
+}
+
+// if needed, create a macro that takes a config function to be more flexible than
+// register_component_config.
```
lockbox: what is this comment in reference to? is this supposed to be attached to `register_component_config`?

incubation/styx-uconf/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+//
+// Until we can fix that we may have do have a structure like we have here.
+#[derive(Deserialize, Debug)]
+pub struct UnifiedConfig {
+    pub processors: Vec<ProcessorConfig>,
+    pub version: u16,
+}
+
+#[derive(Deserialize, Debug)]
+pub struct ProcessorConfig {
```
lockbox, line 146: should we implement `Into<ProcessorBuilder> for ProcessorConfig`? or is that too much magic?

incubation/styx-uconf/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+#[derive(Deserialize, Debug)]
+#[serde(transparent)]
+pub struct ComponentConfig {
+    pub config: serde_yaml::Value,
+}
+
+/// Combine a processor config with the registered components store.
+///
+/// Users should prefer [realize_unified()] or the styx-uconf-cli as they are simpler.
+pub fn realize_processor_config(
```
lockbox, line 221: this is essentially `Into<>` but needs an input, so maybe the into wont work here anyways

incubation/styx-uconf/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+        .processors
+        .iter()
+        .map(|proc_config| realize_processor_config(proc_config, mapper))
+        .collect()
+}
+
+/// Realize a [`ProcessorBuilder`] list from a yaml unified config.
+///
+/// See the crate level documentation for more information.
+pub fn realize_unified(
```
lockbox, line 311: these functions and doc comments feel off-by-one to me, where this is obviously the top-level entry, should this comment then be on the above function that is returning the `Vec<ProcessorBuilder>`? (this i guess does at the moment, but this top level signature will change when we add the connections, and if we add workspaces etc.)

no action needed? idk more a passing comment than anything as i reread this.

incubation/styx-uconf/src/components.rs
YurBoiRene: I actually did this in the upcoming gdb pr so I will just remove

incubation/styx-uconf/src/lib.rs
YurBoiRene: No, because it requires the `ProcessorComponentsStore` to "realize" it

incubation/styx-uconf/src/lib.rs
YurBoiRene: Yeah `unified_config()` is top level. I referred the reader to the module level documentation just because I didn't want to copy paste the examples/docs only to have to modify them when we change it with the workspaces, etc.

PR #66, opened 2025-08-22
docs: Updated README.md and CONTRIBUTING.md, added CODE_OF_CONDUCT.md

kudu-emilie:


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,39 @@
...
+* We will exclude you from interaction if you insult, demean or harass anyone. That is not welcome behavior. We interpret the term “harassment” as including the definition in the [Citizen Code of Conduct](https://github.com/stumpsyn/policies/blob/master/citizen_code_of_conduct.md); if you have any lack of clarity about what might be included in that concept, please read their definition. In particular, we don’t tolerate behavior that excludes people in socially marginalized groups.
+* Private harassment is also unacceptable. No matter who you are, if you feel you have been or are being harassed or made uncomfortable by a community member, please contact the maintainers immediately. Whether you’re a regular contributor or a newcomer, we care about making this community a safe place for you and we’ve got your back.
+* Do not make casual mention of slavery or indentured servitude and/or false comparisons of one's occupation or situation to slavery. Please consider using or asking about alternate terminology when referring to such metaphors in technology.
```
YurBoiRene: we use master-slave terminology in our peripherals. not an issue I think but just making you aware.

CODE_OF_CONDUCT.md
YurBoiRene, line 1: This is good. I thin the only thing it needs is a "Kudu Dynamics LLC holds the final say in any Code of Conduct violations"

CONTRIBUTING.md
```diff
@@ -30,6 +30,15 @@ If something is more clunky than you would like, or is missing functionality tha
 would like, please let us know. Additionally if there is already an issue for it
 please comment +1 on the issue so we can track interest in things and adjust priority.
 
+#### Note for Maintainers
+
+Once a feature is selected for implementation, define sub-tasks for the feature as checkboxes
+in the issue description. Generally the feature should be scoped to be reviewable as one pull request.
+If the feature is sufficiently large and warrants multiple self-contained pull requests, capture those
+tasks as separate issues and apply an approriate Area (A-*) label to group the issues together.
+We uses [Bevy's](https://bevy.org/learn/contribute/reference/triage/) system of Triaging to label our issues.
```
YurBoiRene: I would say we are inspired from Bevy's system. Realistically we should (and have) changed it to meet our needs. E.g. exclusion on "controversial" tags and priority tags

CONTRIBUTING.md
```diff
@@ -30,6 +30,15 @@ If something is more clunky than you would like, or is missing functionality tha
 would like, please let us know. Additionally if there is already an issue for it
 please comment +1 on the issue so we can track interest in things and adjust priority.
 
+#### Note for Maintainers
```
YurBoiRene: Change h4 to "Project Issue Labels and Organization"

README.md
```diff
@@ -7,6 +7,8 @@ of rapid emulator development for debugging target systems. The Styx Emulator
 is primarily a Rust library with bindings in multiple languages and
 extensions for multiple tools.
 
+Styx is under active development and does not currently follow Semantic Versioning. Beware of breaking changes!
```
YurBoiRene, line 10: :100:

CODE_OF_CONDUCT.md
kudu-emilie: Ah well should probably change that terminology then. What should we replace it with?
Controller/Worker
Leader/Follower
Other?

lockbox: I think he means it in the components of the protocol that are called the master or slave nodes/addresses etc. Which we should keep since in the protocols we are modeling that is the protocol level term, and changing that would only confuse people and lead to down-the-road bugs etc. But yea absolutely dont use that verbiage when doing other stuff 100%.

YurBoiRene: Yeah, just thought I should mention. When we're interfacing with existing processor docs and protocols we should keep the terminology. Otherwise definitely more confusing.

kudu-emilie: Right - I was asking what we should change it to in the code, but if you're both saying that's inadvisable atm I won't push. Though we should explore what it would take to change it in the future. Enough projects have moved away from that language that we should see what the new consensus is and conform to that, even if it might introduce slight confusion.  Let's circle back when we need to add a new processor/protocol that uses that old language. 

We should put a note in the Code of Conduct to clarify that this language is currently necessary in order to support existing protocols/processors. But that it should not be used out of context.

PR #95, opened 2025-09-02
Unified Configuration GDB Plugin Support

### LINKED ISSUE #91: GDB Support in Unified Configuration

YurBoiRene:
blocked by #44 

closes #91


examples/styx-uconf-gdb/styx.yaml
```diff
@@ -0,0 +1,19 @@
+version: 1
+processors:
+  - name: FreeRTOS Test Processor
+    processor: ppc_4xx
+    backend: Pcode
+    executor:
+      id: gdb
+      config:
+        connection: 127.0.0.1:9999
+        arch: Ppc405
```
lockbox: question about the discriminant here being `Arch`, wouldn't it have to be a variant?

edit: looks like its a variant internally, could we change it to `arch_variant` just for consistency?

YurBoiRene: good catch

PR #96, opened 2025-09-04
PCS and Experimental Processor Config

YurBoiRene:
## Peripheral Communication Service

This is the PCS as detailed in RFC-03. It is by no means complete but I need an MVP.

### DIFF
--- a/incubation/styx-pcs/src/lib.rs
+++ b/incubation/styx-pcs/src/lib.rs
@@ -1,0 +1,84 @@
+// SPDX-License-Identifier: BSD-2-Clause
+//! Styx Peripheral Communication Service
+//!
+//! The Peripheral Communication Service (PCS) enabled multiple gRPC servers (processors) to communicate
+//! with each other by spawning proxies to communicate.
+//!
+//! Additional proxy services can be implemented by registering a new [`ProxyService`].
+//!
+//! Start the PCS using [`start_pcs()`].
+//!
+//! ## Components
+//!
+//! Components are a concept borrowed from styx-uconf.
+//! Basically, a component is a const object (usually a function pointer) that is collected via inventory.
+//! Collection via inventory means that any crate being compiled can register a component.
+//! All registered components are collected by the crate that defined the component type.
+//!
+//! ## Proxy Service
+//!
+//! styx-pcs defines the [`ProxyService`] component.
+//! A Proxy service proxies traffic between several processors (gRPC servers) over a
+//! peripheral protocol (e.g. uart, spit, i2c, etc.).
+//!
+//! This crate currently defines a `uart` proxy service.
+//! Users can define their own by defining a function with the [`ProxyService`] signature and calling
+//! [`inventory::submit!()`] on the component instantiated using [`component!()`].
+//!
+//! ## Unimplemented
+//!
+//! - Spawn clients from styx-devices
+//!   - This is present in the config but not implemented yet
+//!
+
+/// Component logic for register peripheral implementations.
+pub mod components;
+use components::ComponentReference;
+
+/// Configuration of the PCS, should be deserializable.
+mod config;
+pub use config::*;
+
+/// Included peripheral implmentations.
+mod peripherals;
+pub use peripherals::*;
+
+/// Processors that can be configured to connect to.
+mod processor;
+pub use processor::*;
+
+use styx_core::prelude::*;
+use tokio::runtime::Handle;
+
+/// Spawns the Peripheral Component Service on the given runtime
+pub fn start_pcs(config: &PcsConfig, runtime: &Handle) -> Result<(), UnknownError> {
+    // TODO we should spawn local devices below
+    // create a list of remote devices (processors/gRPC servers) and a list of local devices
+    let (remote_devices, _spawn_devices) = config.devices.separate();
+    let processors = Processors::from_config(remote_devices)?;
+    let peripherals =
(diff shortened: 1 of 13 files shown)


incubation/styx-pcs/src/peripherals/uart.rs
```diff
@@ -0,0 +1,159 @@
+// SPDX-License-Identifier: BSD-2-Clause
```
lockbox, line 1: this really makes me feel like this should be either importable, or used as like the foundation as the model of the respective peripheral etc. since its essentially the same. then the actual peripheral stuff people would have to implement would just be the state machines and hooks etc.

incubation/styx-pcs/tests/uart.rs
```diff
@@ -0,0 +1,280 @@
...
+    let yaml = format!(
+        r#"
+        devices:
+            - !Remote
+              id: sender
+              endpoint: http://127.0.0.1:{port_send}
+            - !Remote
+              id: echo
+              endpoint: http://127.0.0.1:{port_echo}
+        proxies:
```
lockbox: i dont know what, but i think that a name other than "proxies" would be better here (protocols? peripherals? connections?), idk, personally im a fan of connections now that i said it out loud

incubation/styx-pcs/src/lib.rs
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: BSD-2-Clause
+//! Styx Peripheral Communication Service
```
lockbox, line 2: This crate is magic, and probably requires a nice detailed diagram / flow chart of how things fit together. also does this support spawning things from `styx-devices`?

styx/core/styx-grpc/src/lib.rs
```diff
@@ -27,6 +27,23 @@ pub trait Validator: Send + Sync + 'static {
...
 pub mod io {
     pub mod uart {
+
+        #[derive(Debug)]
+        pub enum SubscribeDirection {
+            Rx = 0,
+            Tx = 1,
+            Both = 2,
+        }
+        impl From<Option<subscribe_request::Direction>> for SubscribeDirection {
```
lockbox: if this is from `Option<subscribe_request::Direction>` this should be a `TryFrom<T>` instead of a `From<T>` since it can fail (and will panic in the `value.unwrap()` as currently written)

styx/core/styx-processor/src/processor/config/mod.rs
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: BSD-2-Clause
+pub mod core_configs;
```
lockbox: is there a nice little example of tossing things into config values? (eg. using some of the `core_config` stuff in the processor). i think that would be a great tangible example / user guide to make (probably also for a docs page once this is live?)

lockbox (commented):
Looking good 😎 , mostly just comments about documenting the magic. gettin real close

incubation/styx-pcs/tests/uart.rs
YurBoiRene: Damn I changed this from connections to proxies recently haha. I agree though, connections is good

lockbox: Proxies could be the list of pcs' things are going to connect to (for now single node, eventually a fallback etc)

incubation/styx-pcs/src/peripherals/uart.rs
YurBoiRene: Ah, if I understand correctly, you're saying that most of this code will be reused by other peripherals (mainly the gRPC client implementation) so that should be taken out and provide an interface for implementers to work off of to implement their own peripherals?

Btw, the uart is "importable" in the sense that any PCS config can use the `uart` connections/proxy service.

styx/core/styx-processor/src/processor/config/mod.rs
YurBoiRene: I removed the config stuff, since the PCS doesn't need it at the moment.

incubation/styx-pcs/src/lib.rs
YurBoiRene: Added some more documentation

incubation/styx-pcs/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+//! Collection via inventory means that any crate being compiled can register a component.
+//! All registered components are collected by the crate that defined the component type.
+//!
+//! ## Proxy Service
+//!
+//! styx-pcs defines the [`ProxyService`] component.
+//! A Proxy service proxies traffic between several processors (gRPC servers) over a
+//! peripheral protocol (e.g. uart, spit, i2c, etc.).
+//!
+//! This crate current defines a `uart` proxy service.
```
lockbox: s/current/currently/?

incubation/styx-pcs/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+//! Users can define their own by defining a functino with the [`ProxyService`] signature and calling
```
lockbox: s/functino/function/

incubation/styx-pcs/src/peripherals/uart.rs
lockbox: i was coming at it more from the "this logic looks very close to what some clients / generic server stubs look like, we might want to reuse stuff if possible."

PR #104, opened 2025-09-23
Hexagon ISA Execution Support

yuv418:
- [x] clean up current stuff
- [x] predicate anding
- [x] add RFC to this
- [x] decouple fetch/execute pcode from `PcodeBackend`
- [x] extract common functionality in `PcodeBackend` for implementing `CpuBackend` so it can be shared with `HexagonPcodeBackend`


styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/pkt_semantics.rs
```diff
@@ -0,0 +1,45 @@
...
+        // Should I be unwrapping?
+        // Not happy about clones
+        let mut input = inputs[0].clone();
+        input.offset += DEST_REG_OFFSET;
+        let reg_val = backend.read(&input).unwrap();
+
+        // For now, since there are no packet semantics, we should just
+        // use the previously set value.
+        //
+        // TODO: update when packet semantics come into play
```
YurBoiRene: Make `CallOtherHandleError::Other` wrap an `UnknownError` (with #[from]) and then replace unwraps with `err.with_context(|| "blah blah context").

This is the newish standard way of handling unrecoverable errors in styx.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/tests/branching.rs
```diff
@@ -0,0 +1,141 @@
...
+    assert_eq!(r0, 929);
+    assert_eq!(r4, 32);
+    assert_eq!(r5, r4);
+    assert_eq!(r3, 33);
+    assert_eq!(r2, 34);
+}
+
+// TODO: jumpr branch (indirect), conditional branch,
+// and a branch that isn't at the end of the packet!
+// FIXME: come back to do this
```
YurBoiRene: I think this can be removed unless you see something that should be tested that isn't

styx/core/styx-cpu-pcode-backend/styx-sleigh-bindings/sleigh/sleigh.cc
```diff
@@ -658,6 +658,20 @@ void Sleigh::resolve(ParserContext &pos) const
   pos.setParserState(ParserContext::disassembly);
 }
 
+void Sleigh::setContextVariableCached(const string &nm,const Address &addrLo,
```
YurBoiRene, line 664: Move this to bridge.cc.

It will get nuked when we update ghidra sleigh code in the future and will cause a headache

styx/core/styx-cpu-type/src/arch/hexagon/variants.rs
```diff
@@ -0,0 +1,352 @@
...
+            }
+
+            fn gdb_target_description(&self) -> crate::arch::GdbTargetDescriptionImpl {
+                <$target_description>::default().into()
+            }
+        }
+    };
+}
+
+// TODO: change to HexagonGeneralRegistersWithHvx? Need to find out which DSPs have HVX and which don't.
```
YurBoiRene: from my reading of [llvm's hexagon tablegen file](https://github.com/llvm/llvm-project/blob/2802ab673e1613bd158bd2a0998c7604001fc7fb/llvm/lib/Target/Hexagon/Hexagon.td), all versions v60 and beyond have some form of HVX.

styx/core/styx-cpu-type/src/arch/hexagon/gdb_targets.rs
```diff
@@ -0,0 +1,157 @@
...
+        (57, HexagonRegister::C25.register()),
+        (58, HexagonRegister::C26.register()),
+        (59, HexagonRegister::C27.register()),
+        (60, HexagonRegister::C28.register()),
+        (61, HexagonRegister::C29.register()),
+        (62, HexagonRegister::UtimerLo.register()),
+        (63, HexagonRegister::UtimerHi.register())
+    ]);
+
+    // TODO: which variants have HVX, and which ones don't?
```
YurBoiRene: left a comment in variants.rs, I think revisions v60+ have HVX

styx/core/styx-cpu-type/src/arch/hexagon/gdb_targets.rs
```diff
@@ -0,0 +1,157 @@
...
+    pub static ref HEXAGON_CORE_HVX_CPU_REGISTER_MAP: BTreeMap<usize, CpuRegister> = HEXAGON_CORE_CPU_REGISTER_MAP
+        .clone().into_iter().chain(HEXAGON_CORE_HVX_CPU_REGISTER_MAP.clone().into_iter()).collect();
+}
+
+#[gdb_target_description]
+#[derive(Debug, Default)]
+pub struct HexagonCpuTargetDescription {
+    #[args(
+        // TODO: what should this be?
+        gdb_arch_name("hexagon-core"),
```
YurBoiRene, line 241: I think this should be `hexagon`

styx/core/styx-cpu-type/src/arch/hexagon/gdb_targets.rs
```diff
@@ -0,0 +1,157 @@
...
+    )]
+    args: PhantomData<()>,
+}
+
+#[gdb_target_description]
+#[derive(Debug, Default)]
+pub struct HexagonHvxCpuTargetDescription {
+    #[args(
+        // TODO: what should this be?
+        gdb_arch_name("hexagon-hvx"),
```
YurBoiRene, line 254: I think this should be `hexagon:hvx`

styx/core/styx-cpu-pcode-backend/styx-pcode-translator/src/translator.rs
```diff
@@ -270,14 +331,185 @@ mod arm_tests {
...
+        )
+        .unwrap();
+        translator.set_context_option(ContextOption::ThumbMode(true));
+
+        let mut pcodes = Vec::new();
+        let result = translator.get_pcode(start, &mut pcodes, ());
+        assert!(matches!(result, Err(SleighTranslateError::BadDataError)));
+    }*/
+}
+
```
YurBoiRene: I think you can remove?

styx/core/styx-cpu-pcode-backend/styx-pcode-translator/src/translator.rs
```diff
@@ -270,14 +331,185 @@ mod arm_tests {
...
+        // This is 0x12345678 in hex, for now, because
+        // I can't figure out how to write an immediate in hex
+
+        // The first four bytes used in immediate would be to set the context register
+        // then the next four are for the actual instruction
+        let (start, mut translator) = manual_test_decompile("{ R0 = add(R0, #305419896); }");
+        translator.set_context_option(&ContextOption::HexagonImmext(0xffffffff));
+
+        // TODO: assert that immext is right
+        // The first get_pcode call will set immext, and won't return any pcodes
```
YurBoiRene: TODO?

styx/core/styx-cpu-pcode-backend/styx-pcode-translator/src/translator.rs
```diff
@@ -270,14 +331,185 @@ mod arm_tests {
...
+
+    // Hexagon duplex instructions basically encode 2 instructions in 32 bits
+    #[cfg_attr(miri, ignore)]
+    #[test]
+    fn test_decompile_hexagon_duplex() {
+        let (start, mut translator) = manual_test_decompile("{ R2 = R3; R3 = R2; }");
+        translator.set_context_option(&ContextOption::HexagonImmext(0xffffffff));
+        translator.set_context_option(&ContextOption::HexagonSubinsn(1));
+
+        // TODO: assert that immext is right
```
YurBoiRene: TODO?

styx/core/styx-cpu-pcode-backend/styx-pcode-translator/src/translator.rs
```diff
@@ -270,14 +331,185 @@ mod arm_tests {
...
+        translator.set_context_option(&ContextOption::HexagonImmext(0xffffffff));
+
+        let mut pcodes = Vec::new();
+        let bytes_used = translator.get_pcode(start, &mut pcodes, ()).unwrap();
+
+        assert_eq!(bytes_used, expected_bytes_used);
+        assert_eq!(pcodes.len(), expected_pcodes_len);
+    }
+
+    // TODO: fix this
```
YurBoiRene: what needs fixing?

styx/core/styx-cpu-pcode-backend/styx-pcode-translator/src/translator.rs
```diff
@@ -270,14 +331,185 @@ mod arm_tests {
...
+                &ArchVariant::Hexagon(HexagonMetaVariants::QDSP6V66(QDSP6V66 {})),
+                loader,
+            )
+            .unwrap(),
+        )
+    }
+
+    fn auto_test_decompile(asm: &str, expected_bytes_used: u64, expected_pcodes_len: usize) {
+        let (start, mut translator) = manual_test_decompile(asm);
+        // TODO: how to pass context options from the individual tests?
```
YurBoiRene: Could take a ref to a context option slice? Do you still need this?

styx/core/styx-cpu-pcode-backend/styx-pcode-translator/src/translator.rs
YurBoiRene, line 1: In general, I am okay if you don't spend too much time making hexagon tests here unless there are specific issues in the sla that are prone to breaking. Running firmware will be a much more comprehensive and time-efficient test.

styx/core/styx-cpu-pcode-backend/styx-pcode-sleigh-backend/src/context_internal.rs
```diff
@@ -10,11 +10,21 @@ pub struct ContextInternal {
 }
 
 impl ContextInternal {
-    pub fn set_variable_default(&mut self, name: impl AsRef<str>, value: u32) {
+    pub fn _set_variable(
```
YurBoiRene: Looks like this isn't used from what I can see. If planned for future use document what that is, otherwise feel free to delete.

styx/core/styx-cpu-pcode-backend/styx-pcode-sleigh-backend/src/sleigh.rs
```diff
@@ -162,8 +163,16 @@ impl<L> Sleigh<L> {
...
-        self._context.set_variable_default(variable, value);
+    pub fn set_variable(&mut self, variable: &str, value: u32) {
+        let space_manager: &ffi::AddrSpaceManager = self.obj.upcast_ref();
+        let default_code_space = space_manager.getDefaultCodeSpace();
+
+        cxx::let_cxx_string!(variable_cxx = variable);
+        let sleigh: Pin<&mut ffi::Sleigh> = self.obj.as_mut();
+        // safety: this should get dropped?
+        let addr_lo = unsafe { ffi::new_address(default_code_space, 0) };
+        let addr_hi = unsafe { ffi::new_address(default_code_space, u64::MAX) };
```
YurBoiRene: I assume you are asking if `addr_lo` and `addr_hi` should be dropped? They are `UniquePtr<>` which wraps c++ `std::unique_ptr` which handles the memory dealloc for us. I.e. they will get dropped/dealloc'd when they go out of scope like Box.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/execution_helper.rs
```diff
@@ -0,0 +1,550 @@
...
+                        Ok(PktState::InsidePacket(insn_array))
+                    }
+                    // We're the first in a packet. There is no Parse sequence
+                    // to indicate the start of a packet, so we must determine this
+                    // by looking at the previous packet's state.
+                    //
+                    // (see section 10.5 for context)
+                    PktState::PktEnded(_) => Ok(PktState::PktStarted(insn_array)),
+
+                    _ => unreachable!("invalid packet sequence"),
```
YurBoiRene, line 351: nit: AIUI, this is can be triggered by malformed code so if we can avoid panic we should. Also add some simple context like "Packet loop bits indicated NotEndOfPacket but previous packet state is invalid: {}".

If in fact this is not able to be triggered by malformed code then ignore me, unreachable is correct.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/execution_helper.rs
```diff
@@ -0,0 +1,550 @@
...
+        prev_state: &PktState,
+        pc: u32,
+    ) -> Result<PktState, HexagonFetchDecodeError> {
+        // Duplex always ends the packet - section 10.3,
+        // duplexes are in slots 0 and 1, which are the last
+        // 2 slots in a packet.
+        //
+        // As such, we can conclude that the second sub-instruction
+        // in a duplex is the last instruction in a packet.
+        match prev_state {
```
YurBoiRene: nit: could use `if matches!(...) {}`

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/execution_helper.rs
```diff
@@ -0,0 +1,550 @@
...
+
+        // Get the PC
+        self.pc_varnode.offset = pc as u64;
+        trace!("got pc is {pc}");
+
+        // We read four instructions, since some decoding requires looking
+        // ahead. Four instructions because every packet in Hexagon is at most
+        // four instructions.
+        let insn_data_wide = mmu
+            .read_u128_le_virt_code(self.pc_varnode.offset, backend)
```
YurBoiRene, line 285: Speaking from experience, I expect this to error in the future.

If the PC executes a <4 insn packet at the edge of a memory region then this will read out of bounds and fail. 

E.g. if a 0x100 sized region is defined at 0x0 and the PC points to a single instruction at 0xFC then this will error as "counldn't fetch".

I had to debug this in the pcode translator callback code because ghidra wants a 16 byte buffer filled for instruction translation.

I am not sure if we will ever encounter this in emulation so I would add a nice comment and a friendly message in the `with_context` call to warn any future hexites (name I just made up. what I am calling people who want to emulate hexagon).

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/mod.rs
```diff
@@ -0,0 +1,1007 @@
...
+        self.execution_helper = Some(execution_helper);
+
+        Ok(Ok(HexagonFetchDecodeInfo {
+            total_bytes_consumed,
+            ordering,
+        }))
+    }
+}
+
+pub trait HexagonExecutionHelper: derive_more::Debug + Send {
```
YurBoiRene, line 1041: Can you document these methods? When they are called (maybe a description of the execution cycle) and what their arguments mean (e.g. what is instr an option?).

LMK if I missed the documentation somewhere else.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/mod.rs
```diff
@@ -0,0 +1,1007 @@
...
+        let call_other = spec.call_other;
+        let register_manager = spec.register;
+
+        let execution_helper = DefaultHexagonExecutionHelper::default();
+
+        Self {
+            saved_context_opts: SavedContextOpts::default(),
+            regs_written: Vec::with_capacity(10),
+            saved_execution_helper: None,
+            execution_helper: Some(execution_helper), // TODO: performance optimizations
```
YurBoiRene: Curious what this todo means

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/mod.rs
```diff
@@ -0,0 +1,1007 @@
...
+            stop_requested: false,
+            last_was_branch: false,
+            call_other_manager: Some(call_other),
+            saved_reg_context: BTreeMap::new(),
+        }
+    }
+    /// Indicate when we should update the context reg
+    /// and what the new value should be
+    pub fn update_context(&mut self, when: PacketLocation, what: ContextOption) {
+        // TODO: what to do when Now is set outside of prefetch?
```
YurBoiRene: not sure I understand this todo, is it still an issue?

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/mod.rs
```diff
@@ -0,0 +1,1007 @@
...
+                PCodeStateChange::PCodeRelative(offset) => {
+                    // for now assume math is good
+                    let next_index = (i as i64 + offset) as usize;
+                    trace!("Pcode state change relative jump {i}+{offset}={next_index}");
+                    i = next_index;
+                }
+                PCodeStateChange::InstructionAbsolute(new_pc) => {
+                    trace!("Pcode state change absolute jump new PC=0x{new_pc:X}");
+                    self.last_was_branch = true;
+                    // self.set_pc(new_pc)?;
```
YurBoiRene: delete

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/mod.rs
```diff
@@ -0,0 +1,1007 @@
...
+                res.map_err(|e| UnknownError::from_boxed(Box::new(e)))?;
+
+                // This moves from the "now" context option time to the buffer of other already staged
+                // context options from `setup_context_opts`.
+                self.saved_context_opts.get_context_opts()
+            };
+            self.execution_helper = Some(execution_helper);
+
+            // TODO: Optimize?
+            let mut pcodes = vec![];
```
YurBoiRene, line 812: :/ make SmallVec if not a big lift, otherwise, leave as is

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/mod.rs
```diff
@@ -0,0 +1,1007 @@
...
+            self.regs_written.push(regs_in_insn);
+
+            total_bytes_consumed += bytes_consumed;
+        }
+
+        let mut execution_helper = self.execution_helper.take().unwrap();
+        {
+            execution_helper.post_packet_fetch(self);
+
+            // TODO: remove this allocation, and turn this into an option that can be taken and replaced
```
YurBoiRene: is there an allocation here?

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/mod.rs
```diff
@@ -0,0 +1,1007 @@
...
+                            0,
+                            Pcode {
+                                opcode: Opcode::Copy,
+                                inputs: smallvec![VarnodeData {
+                                    space: SpaceName::Register,
+                                    offset: DEST_REG_OFFSET + (*dotnew_regnum + 0x94),
+                                    size: 1
+                                }],
+                                // WARN: is there an issue here with the unique space somehow overlapping?
+                                // WARN: is this too big?
```
YurBoiRene, line 973: shouldn't be an issue. currently spaces are just a large continuous allocation that the Linux memory manager does a good job with only allocating pages we use.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/mod.rs
```diff
@@ -0,0 +1,1007 @@
...
+                        full_pcodes[*i].insert(
+                            0,
+                            Pcode {
+                                opcode: Opcode::Copy,
+                                inputs: smallvec![VarnodeData {
+                                    space: SpaceName::Register,
+                                    offset: DEST_REG_OFFSET + (*dotnew_regnum + 0x94),
+                                    size: 1
+                                }],
+                                // WARN: is there an issue here with the unique space somehow overlapping?
```
YurBoiRene, line 972: possible... not sure about this.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/system/interrupt.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+}
+
+pub fn add_interrupt_callothers<S: SlaUserOps<UserOps: FromStr>>(
+    spec: &mut ArchSpecBuilder<S, HexagonPcodeBackend>,
+) {
+    spec.call_other_manager
+        .add_handler_other_sla(
+            HexagonUserOps::Trap0,
+            InterruptGenericStub { from: "trap0" },
```
YurBoiRene, line 37: Image

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/tests/dotnew.rs
```diff
@@ -0,0 +1,1018 @@
...
+            .unwrap();
+        cpu.write_register(HexagonRegister::R10, mem_start as u32)
+            .unwrap();
+
+        let report = cpu.execute(&mut mmu, &mut ev, 1).unwrap();
+        assert_eq!(
+            report.exit_reason,
+            TargetExitReason::InstructionCountComplete
+        );
+        /*for typ in test.types {
```
YurBoiRene: should this be removed

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/tests/duplex.rs
```diff
@@ -0,0 +1,49 @@
...
+    let r2 = cpu.read_register::<u32>(HexagonRegister::R2).unwrap();
+    let r3 = cpu.read_register::<u32>(HexagonRegister::R3).unwrap();
+    let r4 = cpu.read_register::<u32>(HexagonRegister::R4).unwrap();
+
+    assert_eq!(r2, 1905856528u32);
+    assert_eq!(r3, 100);
+    assert_eq!(r4, 100 * 470);
+}
+
+// FIXME: test
```
YurBoiRene: ?

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/tests/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+mod reg_postfix;
+mod regpair;
+mod sequencing;
+
+pub fn setup_asm(
+    asm_str: &str,
+    expected_asm: Option<Vec<u8>>,
+) -> (HexagonPcodeBackend, Mmu, EventController) {
+    styx_util::logging::init_logging();
+    // objdump from example ppc program
```
YurBoiRene: I assume this is left over copy paste?

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/tests/regpair.rs
```diff
@@ -0,0 +1,76 @@
+// SPDX-License-Identifier: BSD-2-Clause
+use crate::arch_spec::hexagon::regpairs::REGPAIR_MAP;
+use crate::arch_spec::hexagon::tests::*;
+
+#[test]
+fn verify_regpairs() {
+    // make sure the mapping is acceptable by going through
+    // the styx sla mapping and making sure they both align
+    // TODO: not clear how to do this, but maybe make sure that every
```
YurBoiRene: what is needed to make this test fufill this requirement?

YurBoiRene (commented):
Lots of good stuff here, great work! I appreciate the additional comments/documentation.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/mod.rs
yuv418: it is kind of a big lift, because all the pcode generation stuff (including in the translator and maybe even the bridge) takes a Vec. I think a better solution is to allocate these vecs beforehand and clear/reuse them. If that's too much in your opinion, i'll just leave it as is

styx/core/styx-cpu-type/src/arch/hexagon/gdb_targets.rs
yuv418: https://github.com/bminor/binutils-gdb/blob/f86c456bec7969779dd044c313e419f8a177f6eb/gdb/features/rs6000/powerpc-32.xml#L13 some other arches seem to show `name:core`, is it possible it would be `hexagon:core`? I don't think GDB actually has an architectural definition in this format for hexagon, making me wonder if it even has full support for hexagon.

styx/core/styx-cpu-pcode-backend/styx-pcode-translator/src/translator.rs
yuv418: this would require reading context options, which is a whole separate deal that isn't really necessary at this point, so i'm not going to implement htis

styx/core/styx-cpu-pcode-backend/styx-pcode-translator/src/translator.rs
yuv418: i deleted this test, it's pointless with the hexagon backend

styx/core/styx-cpu-pcode-backend/styx-pcode-sleigh-backend/src/sleigh.rs
yuv418: yeah, i eventually realized this and forgot to delete the comment. yay smart pointers

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/mod.rs
yuv418: this comment is quite likely from struct member that was deleted in a previous refactor. I deleted this comment

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/mod.rs
yuv418: I think I moved that allocation (the ordering) somewhere else and made it a smallvec, so this isn't a huge deal anymore IIUC. Removing comment.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/tests/duplex.rs
yuv418: yeah no clue. looks fine, removing.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/tests/mod.rs
yuv418: yes, also big endian should be little endian (a few lines down), but I guess keystone doesn't change its output either way. fixed all of that

styx/core/styx-cpu-pcode-backend/styx-pcode-translator/src/translator.rs
yuv418: Can I just remove all of the stuff in `translator.rs`? At this point this stuff is covered by the more comprehensive test suite.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/mod.rs
yuv418: Yeah, I think I can just copy some stuff from a prior RFC in a GitHub comment somewhere for this. Sorry, missed this one.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/tests/dotnew.rs
yuv418: depends if I want to put more stuff in the execution report

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/tests/regpair.rs
yuv418: I'm pretty sure whatever is here now is an acceptable way of verifying things are correct

yuv418: I think maybe what I had in mind was maybe to access the hashmap that `get_register_rev`  and find all registers > 32 bits, and then see if those have an appropriate mapping in this hand-defined register pair map that I have? It doesn't really seem necessary at this point, because I had implemented the register pair handling with `RegisterManager` previously and realized that won't work when executing code, so I switched to using whatever offsets for registers were programmed into the slaspec.

styx/core/styx-cpu-pcode-backend/styx-pcode-translator/src/translator.rs
yuv418: see comment below - am thinking we can just remove all of this

styx/core/styx-cpu-pcode-backend/styx-pcode-translator/src/translator.rs
yuv418: see comment below - am thinking we can just remove all of this

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/execution_helper.rs
yuv418: I added a comment that this might error out if we are at the end of the memory region that has our code, and added a bit about that to the error message if this call fails.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/mod.rs
yuv418: Done!

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/tests/dotnew.rs
yuv418: I have concluded this is not worth doing. I'm pretty sure if whatever I was testing here failed, the test will crash or panic or something

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/execution_helper.rs
yuv418: The list of states that aren't handled are:

Standalone packet (pkt with 1 instruction) - this will never be the `prev_state` since a standalone packet is considered the "end of a packet," the `fetch_decode_packet` function actually returns after seeing a standalone and the "next packet" will show up as a `PktEnded`. In other words, the standalone packet will get "converted" to have the state `PktEnded` by the time this function is called.
First duplex or duplex at start of packet - there's a fast path where the second duplex does not require any of this fetching because we already know what to do for the second duplex. the second duplex sets the packet state to `PktEnded`. 

These cases should from the parsing logic always be handled earlier, even in the case of malformed input.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/mod.rs
yuv418: I think this has to do with semantics I decided on documented elsewhere. will update in a bit

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/pkt_semantics.rs
yuv418: Is this something I should do for all unwraps in my code? Either way, I fixed it here.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/execution_helper.rs
```diff
@@ -0,0 +1,553 @@
...
+
+        trace!("fetching 4 instruction words from memory");
+
+        // Get the PC
+        self.pc_varnode.offset = pc as u64;
+        trace!("got pc is {pc}");
+
+        // We read four instructions, since some decoding requires looking
+        // ahead. Four instructions because every packet in Hexagon is at most
+        // four instructions.
```
androm3da, line 280: It's probably more accurate to say that packets can contain as many as five instructions among four instruction words.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/dotnew.rs
```diff
@@ -0,0 +1,157 @@
...
+    #[bits(15..=16, r)]
+    nv_reg_offset: u2,
+}
+
+/// Check hexagon instruction to see if it references a dot-new register.
+/// This does _not_ refer to section 6.1.4 "Dot-new predicates," but rather
+/// section 5.6 regarding new-value stores.
+///
+/// Since dot-new registers are encoded using offsets from the producing location,
+/// (eg. the new-value register to be used here is the output register from 2 instructions
```
androm3da, line 85: Note that in the case of HVX dot new references, it's the distance in *HVX* instruction words.  Therefore you exclude the scalars when counting the offset for HVX.

Not sure if you support HVX yet but it's a subtlety that might be easy to get wrong and not notice until you find an interesting packet.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/execution_helper.rs
androm3da: Actually - no, that's wrong.  You're still limited to the four execution slots when duplexing.  I was thinking of how the assembler keeps state for the instructions in memory before encoding and in that case constant extenders show up as their own "instruction".

Please disregard.

lockbox:
@androm3da thanks for the sanity check here, a lot of this is educated guesstimation 🙏

androm3da:
> @androm3da thanks for the sanity check here, a lot of this is educated guesstimation 🙏

I haven't done an exhaustive look yet but so far, so good from what I see.

BTW -- we have recently published [a repo with some tests for QEMU hexagon](https://github.com/qualcomm/qemu-hexagon-testing).  It's not much beyond what is already upstream and in our public `hex-next` branch.  But in the coming ~weeks we'll also add our tool that we've used for verification of QEMU hexagon.  It generates some random packets and runs a program in both QEMU and hexagon-sim, comparing the architectural state using the debug interface.  This tool might be useful for you as your implementation matures.

styx/core/styx-cpu-pcode-backend/styx-sleigh-bindings/sleigh/sleigh.cc
yuv418: added docs on this :heavy_check_mark:

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/execution_helper.rs
yuv418: I added a comment on this to make it clear what's happening. Other solutions are too big of a lift or just move the same thing elsewhere.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/mod.rs
yuv418: I added some docs on this

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/mod.rs
yuv418: yeah, uh, idk

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/mod.rs
YurBoiRene: It is good

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/mod.rs
YurBoiRene: Suggestion: Make a new space (`styx_hexagon` maybe) and write intermediate predicate and values there

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/mod.rs
```diff
@@ -0,0 +1,1117 @@
...
+                        // Also, this kind of assumes that the predicate register is either 0x00 or 0xff, and
+                        // nothing else.
+                        full_pcodes[*i].insert(
+                            *ins_loc,
+                            Pcode {
+                                opcode: Opcode::IntAnd,
+                                inputs: smallvec![
+                                    VarnodeData {
+                                        space: SpaceName::Register,
+                                        offset: DEST_REG_OFFSET + (*dotnew_regnum + 0x94),
```
YurBoiRene: What is `0x94` magic number? Probably make a const with documenting comment

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/tests/regpair.rs
YurBoiRene: Okay, you can remove the TODO then

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/pkt_semantics.rs
```diff
@@ -0,0 +1,50 @@
...
+use styx_errors::anyhow::Context;
+use styx_pcode::pcode::VarnodeData;
+use styx_processor::{cpu::CpuBackend, event_controller::EventController, memory::Mmu};
+
+use crate::{
+    call_other::{CallOtherCallback, CallOtherCpu, CallOtherHandleError},
+    PCodeStateChange,
+};
+
+pub const DEST_REG_OFFSET: u64 = 0x600;
```
YurBoiRene, line 14: Can you add a comment here that this is defined in the sleigh specification

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/pkt_semantics.rs
YurBoiRene: As much as possible yes, thanks for fixing :)

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/tests/predicate_anding.rs
```diff
@@ -0,0 +1,458 @@
...
+       0:	08 71 80 14	14807108 { 	p1 = cmp.gt(r0,r1); if (p1.new) jump:t 0x10
+       4:	06 40 00 58	58004006   	jump 0xc
+       8:	01 c3 42 f2	f242c301   	p1 = cmp.gt(r2,r3) }
+       c:	05 c5 00 78	7800c505 { 	r5 = #0x28 }
+      10:	45 cb 00 78	7800cb45 { 	r5 = #0x5a }
+"#,
+    );
+
+    // TODO: use the test case thing to create a couple of cases
+    // where it doesn't pass and where it does
```
YurBoiRene: I think this is done? You use test_case here

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/mod.rs
yuv418: It's the predicate register, but I shouldn't do it like this. I added a method somewhere to retrieve this constant, so let me try to use that

styx/core/styx-cpu-type/src/arch/hexagon/gdb_targets.rs
YurBoiRene: I don't think it does..... not sure why we didn't check that

lockbox: Yeah, the gdb situation is an upstream affair unfortunately...if there is a hexagon gdb fork we should attempt to match the format of any existing thing else users will be presented with either inscrutable gdb client errors, or a gdb client that aborts. 

llvm/lldb support hexagon, from [these slides](https://llvm.org/devmtg/2013-11/slides/Riley-DebugginWithLLDB.pdf) it looks like they did a lot of the work a gdbserver would need to do so its probably worth seeing what the api/specification/format of the xml schema they use for hexagon

yuv418: I was trying to find what lldb might do for this to no luck after a lazy search. Let me see if I can find it again.

lockbox: looks like `hexagon-core` is what we want: https://lldb.llvm.org/cpp_reference/ABISysV__hexagon_8cpp_source.html

lockbox: also look at the proper organization for register numbers ^

yuv418: llvm has more detailed register number mappings for dwarf (eg for system registers), so I can try to update and add those. what should I do for hvx?

lockbox: aiui only llvm/lldb has support for hvx. you can add a `hexagon-hvx` version now, or we can do it later.

register list: https://github.com/llvm/llvm-project/blob/17efa572c3827fe12e37d0f1f28d20a8e44890c7/clang/lib/Basic/Targets/Hexagon.cpp#L180
example hvx xml compatible with lldb: https://gitlab.com/qemu-project/qemu/-/blob/master/gdb-xml/hexagon-hvx.xml
example haxagon-core xml compatible with lldb+gdb: https://gitlab.com/qemu-project/qemu/-/blob/master/gdb-xml/hexagon-core.xml
MR to qemu implementing the above: https://patchew.org/QEMU/cover.1683214375.git.quic._5Fmathbern@quicinc.com/

yuv418: I'll add it now, and update the register mappings. there should be a dwarf mapping file i added to the slaspec directory that has this stuff too based on the mappings in the hexagon.cpp that you linked

yuv418: https://github.com/llvm/llvm-project/blob/main/lldb/source/Plugins/ABI/Hexagon/ABISysV_hexagon.cpp

yuv418: Okay, this was a bit more complicated to find than I thought. I've gone ahead and updated the mappings, let me know if there are problems. 

Suppose we have 32 general purpose regs, and 16 system registers in a made-up ISA. 

There are separate arrays of core/system registers (this split is called a GDB feature), which are read in from the XML files. The ordering of registers for each type is generated from a Python script. You can generate this by running `python scripts/feature_to_c.py gdb-xml/hexagon-*`.  If you run this, it generates an array of `GDBFeature`. Each `GDBFeature` contains the file name (eg. `hexagon-hvx.xml`), the XML contents, the name of the "gdb feature," an array of the registers defined in that file, and the length of that array.

The backend adds each array of register types to a bigger array (cpu->gdb_regs) in a certain order using `gdb_register_coprocessor`.  Eg. a backend may choose to add `hexagon-core.xml`, `hexagon-sys.xml` and `hexagon-hvx.xml` to the gdb_regs. See https://github.com/quic/qemu/blob/hex-next/target/hexagon/cpu.c#L491, https://github.com/quic/qemu/blob/hex-next/gdbstub/gdbstub.c#L604, and https://github.com/quic/qemu/blob/hex-next/gdbstub/gdbstub.c#L555. 

When QEMU goes to read/write a register at the request of GDB (gdb specifies the register number, which is what we need), it looks at the registers that were added to `cpu->gdb_regs`. So, after the core registers and `gdb_register_coprocessor`, might have general register names [0..31] in an array, and [0, 15] for sys regs (in the made-up ISA). If we requested register 33, we would basically look to see if this is within the first 32 registers (_assume the core registers added first and sys registers added second_) and if not subtract 32 from our register request (so now we are at 1), then look at the second array of sys regs, specifically at the second register in that array. As such, we would get the second register in this [0..15]. See https://github.com/quic/qemu/blob/hex-next/gdbstub/gdbstub.c#L532 for details.

So 0..31 would be mapped to the ordering of registers in the array inside the `GDBFeature`, then 32..47 has the ordering of registers in the systems registers for the corresponding `GDBFeature`, etc. 

To understand where I got the Hexagon mappings from, run the python script mentioned above, then look at the output and look at the arrays for the separate system/hvx/core regs. Also, looks like from https://github.com/quic/qemu/blob/hex-next/target/hexagon/cpu.c#L491 that first we have core (core is added/initialized elsewhere), then HVX, then system registers.

yuv418: Also, the `gdb_arch_name` in Styx seems to be undefined for us. QEMU at least outputs the `<architecture>ARCH_NAME</architecture>` XML tag only if the arch backend defines `gdb_arch_name`, which is not defined for Hexagon. I changed it to `hexagon-core` and `hexagon-hvx`.

See https://github.com/quic/qemu/blob/hex-next/gdbstub/gdbstub.c#L380.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/execution_helper.rs
androm3da: > I am not sure if we will ever encounter this in emulation

Packets that span page boundaries do happen.  Does this correspond to a "memory region"?  If not, it's likely safe to ignore.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/regpairs.rs
```diff
@@ -0,0 +1,514 @@
...
+use crate::register_manager::RegisterCallbackCpu;
+use crate::{
+    arch_spec::ArchSpecBuilder,
+    memory::sized_value::SizedValue,
+    register_manager::{RegisterCallback, RegisterHandleError},
+};
+
+use super::backend::HexagonPcodeBackend;
+
+// TODO: FxHashmap here?
```
lockbox: if these maps are going to be frequently referenced yes please use fxhashmap here

yuv418: theyre only used for testing - can add a comment about that if needed. Eventually, most of the registerhandler stuff in here will be deleted (when we add hvx) support. RegisterHandler doesn't do anything during execution from what I understand

styx/core/styx-cpu-pcode-backend/src/lib.rs
```diff
@@ -152,30 +120,54 @@ fn handle_basic_block_hooks(
 /// processor manual. With no intervention, the instruction will be run again, possibly throwing the
 /// same TLB exception.
 ///
+#[derive(Derivative)]
```
lockbox: This enum definition messes up the doc comment for `PcodeBackend`, please move it so that the doc-comment is anchored properly to the proper type

styx/core/styx-cpu-pcode-backend/src/lib.rs
```diff
@@ -551,7 +525,10 @@ impl CpuBackend for PcodeBackend {
 
     fn set_pc(&mut self, value: u64) -> Result<(), UnknownError> {
         let mut pc_manager = self.pc_manager.take().unwrap();
-        pc_manager.set_internal_pc(value, self);
+        // NOTE: if set_pc starts being called from a branching location,
+        // then we need to be more nuanced here.
+        // TODO: decide whether we should pass from_branch here as well
```
lockbox: is this comment still valid?

styx/core/styx-cpu-pcode-backend/src/lib.rs
```diff
@@ -301,6 +264,77 @@ impl PcodeBackend {
...
+        instruction_pc
+    }
+    fn pre_execute_hooks(
+        &mut self,
+        mmu: &mut Mmu,
+        ev: &mut EventController,
+    ) -> Result<(), UnknownError> {
+        // Trigger code hooks for this address
+        // This is done before pcode generation allowing for the
+        // current instruction to be modified in memory, not sure if this is correct behavior.
```
lockbox, line 297: please elevate this to a doc-comment, and change the speculation into an assertion 👍

styx/core/styx-cpu-pcode-backend/src/get_pcode.rs
```diff
@@ -157,34 +150,66 @@ pub(crate) fn fetch_pcode(
...
+    // attempt to fetch and translate pcodes
+    let result = get_pcode_for_pcode_backend(cpu, pcodes, mmu, ev);
+    // if success, then return early
+    let result_err = match result {
+        Ok(success) => return Ok(success),
+        Err(err) => err,
     };
 
+    // now we know we encountered an error.
+    // this could be an exception that we have to handle or a fatal error
```
lockbox: nit: `This could be an exception that we have to handle, or a fatal error`

styx/core/styx-cpu-pcode-backend/src/get_pcode.rs
```diff
@@ -157,34 +150,66 @@ pub(crate) fn fetch_pcode(
...
+    pcodes: &mut Vec<Pcode>,
+    mmu: &mut Mmu,
+    ev: &mut EventController,
+) -> Result<u64, FetchPcodeError> {
+    // attempt to fetch and translate pcodes
+    let result = get_pcode_for_pcode_backend(cpu, pcodes, mmu, ev);
+    // if success, then return early
+    let result_err = match result {
+        Ok(success) => return Ok(success),
+        Err(err) => err,
```
lockbox: This stuff is a little confusing, could you add to the comment saying something like "return early on success, or save the error to see if we can are able to resolve the target error" or something? Also probably worth adding to the doc-comment that this function has some sort of internal handling (to a degreee) and what the general "error handling" here entails (eg. what limits of standard error handling logic and what is never going to be handled?)

styx/core/styx-cpu-pcode-backend/src/get_pcode.rs
```diff
@@ -69,31 +73,33 @@ impl TryFrom<GetPcodeError> for PcodeFetchException {
     }
 }
 
-/// thin wrapper to [GhidraPcodeGenerator::get_pcode].
-fn get_pcode_at_address(
+/// For use with a PcodeBackend (and not other PcodeBackends). This function wraps
+/// `get_pcode_at_address`, and extracts the needed context options and current PC,
+/// which require `PcodeBackend`-specific function calls.
+fn get_pcode_for_pcode_backend(
```
lockbox, line 79: Since this is only referring to "THE `PcodeBackend`", can you link it a la 

```
[`PcodeBackend`]
```

styx/core/styx-cpu-pcode-backend/src/backend_helper.rs
```diff
@@ -0,0 +1,216 @@
...
+        let space_memory = match space_name {
+            // This is where we define the backing store for each of the spaces added to the
+            // machine, based on their space name. The Ram space is already added above as the
+            // default space and has the [StyxStore] memory storage and the
+            // [SpaceName::Constant] store added by default.
+
+            // Currently this allocates giant vectors which makes space reads/writes very fast
+            // but also theoretically takes a lot of memory. However, Linux's paging system
+            // allows us to allocate lots of memory without actually using any physical memory
+            // until we access it.
```
lockbox: This probably deserves being in a doc-comment, no?

styx/core/styx-cpu-pcode-backend/src/backend_helper.rs
```diff
@@ -0,0 +1,216 @@
...
+    } // no code hook if translate errors, we will catch then on instruction fetch
+    Ok(())
+}
+
+pub struct BackendHelperExecuteInfo<T> {
+    pub report: ExecutionReport,
+    pub execute_single_info: Option<T>,
+}
+
+pub trait BackendHelper<T, Q>: CpuBackend + HasHookManager + Sized {
```
lockbox: This trait should be documented what/why/when if possible

styx/core/styx-cpu-pcode-backend/src/backend_helper.rs
```diff
@@ -0,0 +1,216 @@
...
+    fn execute_single(
+        &mut self,
+        pcodes: &mut Vec<Q>,
+        mmu: &mut Mmu,
+        ev: &mut EventController,
+    ) -> Result<Result<T, TargetExitReason>, UnknownError>;
+    fn set_last_was_branch(&mut self, last_was_branch: bool);
+    fn last_was_branch(&mut self) -> bool;
+
+    fn execute_helper(
```
lockbox, line 185: this having default behavior in a public trait should have doc comment content discussing what it does by default and when you want to override it

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/execution_helper.rs
yuv418: I _think_ this is supposed to be like the entire RAM space, and not corresponding to a page

styx/core/styx-cpu-pcode-backend/src/lib.rs
yuv418: this is in mainline styx, maybe this should be a separate issue https://github.com/styx-emulator/styx-emulator/blob/main/styx/core/styx-cpu-pcode-backend/src/lib.rs#L406? I didn't work on this

styx/core/styx-cpu-pcode-backend/src/get_pcode.rs
yuv418: https://github.com/styx-emulator/styx-emulator/blob/main/styx/core/styx-cpu-pcode-backend/src/get_pcode.rs#L133 also mainline styx, but i can fix this

styx/core/styx-cpu-pcode-backend/src/execute_pcode.rs
```diff
@@ -68,18 +69,48 @@ impl<'a> From<PCodeStateChange> for PCodeStateChangeInner<'a> {
     }
 }
 
-pub fn execute_pcode(
+pub fn execute_pcode<
+    T: CpuBackend
+        + HasSpaceManager
```
lockbox: nit: for this file can we have a new trait / type alias be like `BasePcodeExecutor<T>: CpuBackend + HasSpaceManager + HasHookManager + HasPcodeGenerator<InnerCpuBackend=T> + HasConfig`? That seems to be the common subset here and will reduce the clutter IMO while making it easier to see what dependencies are actually different between the methods etc.

styx/core/styx-cpu-pcode-backend/src/memory/space_manager.rs
```diff
@@ -408,13 +416,16 @@ impl SpaceManager {
...
-    pub fn read_hooked_register(
-        cpu: &mut (impl CpuBackend
-                  + HasSpaceManager
-                  + MmuSpaceOps
-                  + HasHookManager
-                  + HasPcodeGenerator
-                  + HasConfig),
+    pub fn read_hooked_register<
+        B: CpuBackend
+            + HasSpaceManager
```
lockbox: nit: similar to the comment about the `BasePcodeExeutor` trait alias, we could also probably decide to do similar here if you think its worth it.

styx/core/styx-cpu-pcode-backend/src/backend_helper.rs
yuv418: i think this is also from mainline

styx/core/styx-cpu-pcode-backend/src/lib.rs
yuv418: absolutely not. removed.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/pkt_semantics.rs
```diff
@@ -0,0 +1,50 @@
...
+        backend: &mut dyn CallOtherCpu<T>,
+        _mmu: &mut Mmu,
+        _ev: &mut EventController,
+        inputs: &[VarnodeData],
+        output: Option<&VarnodeData>,
+    ) -> Result<PCodeStateChange, CallOtherHandleError> {
+        debug_assert_eq!(inputs.len(), 1);
+        debug_assert!(output.is_some());
+
+        // Should I be unwrapping?
```
YurBoiRene: No more unwrap, can remove this comment.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/pkt_semantics.rs
```diff
@@ -0,0 +1,50 @@
...
+        // Not happy about clones
```
YurBoiRene: To ease your  mind, probably not a big deal. VarnodeData doesn't alloc (afaik). we can perf test later for sure

styx/core/styx-cpu-pcode-backend/src/lib.rs
YurBoiRene: This is correct, from `StyxHook::code()`:
```
    /// This callback is executed before the instruction caught has been
    /// executed by the guest.
```
I probably added that comment before this was decided.

Do you mind removing the self-doubt, e.g. change to just "This is done before pcode generation allowing for the current instruction to be modified in memory"

styx/core/styx-cpu-pcode-backend/src/get_pcode.rs
YurBoiRene: Thanks for fixing

styx/core/styx-cpu-pcode-backend/src/backend_helper.rs
YurBoiRene: Sure, make it a doc comment :)

styx/core/styx-cpu-pcode-backend/src/backend_helper.rs
yuv418: should be ok now

styx/core/styx-cpu-pcode-backend/src/backend_helper.rs
yuv418: should be ok now

styx/core/styx-cpu-pcode-backend/src/lib.rs
yuv418: i just removed this enum since it's not useful anymore with `HexagonPcodeBackend`.

styx/core/styx-cpu-pcode-backend/src/get_pcode.rs
yuv418: done

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/regpairs.rs
yuv418: i added a comment that this is just for testing and removed the todo

styx/core/styx-cpu-pcode-backend/src/lib.rs
yuv418: Opened #122.

styx/core/styx-cpu-pcode-backend/src/backend_helper.rs
yuv418: Not sure what `StyxStore` refers to. I renamed to `BlobStore` for now.

Issue #109, opened 2025-09-25
styx-sla: build panic if diffy encounters CRLF line endings


ntpopgetdope:
**Describe the bug**
When the repo is cloned on an NT host with `git config --global core.autocrlf true` set; building the project (using VSCode Dev Container & WSL2) will fail. The `diffy` crate used to apply patches to SLEIGH defs in `styx-sla` will panic due to: https://github.com/bmwill/diffy/issues/20. The exact point of failure is at [styx/core/styx-cpu-pcode-backend/styx-sla/build.rs#L350-L355](https://github.com/styx-emulator/styx-emulator/blob/main/styx/core/styx-cpu-pcode-backend/styx-sla/build.rs#L350-L355), but subsequent calls to `diffy` functionality will also fail for the same reason e.g. the call to `diffy::apply`.

**To Reproduce**
On a Windows host clone the repo with `autocrlf` enabled & open VSCode to initialise the Dev Container for use:

```shell
git config --global core.autocrlf true
git clone https://github.com/styx-emulator/styx-emulator.git
```
From Dev Container shell; build `styx-cpu-pcode-backend` & observe fail in `styx-sla/build.rs` during `diffy` usage:
```shell
cd styx-emulator/styx/core/styx-cpu-pcode-backend/
export RUST_BACKTRACE=full && cargo build
```

**Expected behavior**
Whilst, Windows hosts are unsupported per https://github.com/styx-emulator/styx-emulator/issues/31, this issue will still be encountered when using WSL2 to build & use Styx. Expected behaviour is failsafe handling of CLRF line endings to workaround issue inherent to `diffy` crate.

**Build Logs**
```shell
warning: unused key `rustdocflags` in [target] config table `cfg(all())`
   Compiling styx-sla v1.2.0 (/workspaces/styx-emulator/styx/core/styx-cpu-pcode-backend/styx-sla)
error: failed to run custom build command for `styx-sla v1.2.0 (/workspaces/styx-emulator/styx/core/styx-cpu-pcode-backend/styx-sla)`
note: To improve backtraces for build dependencies, set the CARGO_PROFILE_DEV_BUILD_OVERRIDE_DEBUG=true environment variable to enable debug information generation.

Caused by:
  process didn't exit successfully: `/workspaces/styx-emulator/target/debug/build/styx-sla-92d183dfa2e31ba5/build-script-build` (exit status: 101)
  --- stdout
  Output Directory:
[...]

PR #110, opened 2025-09-25
fix: styx-sla build panic if diffy encounters CRLF line endings

ntpopgetdope:
naively resolves: https://github.com/styx-emulator/styx-emulator/issues/109
may be useful to cherrypick only line ending normalisation codepath (https://github.com/styx-emulator/styx-emulator/pull/110/commits/ae745f1b654b9c690385630878ea72a274e93c3e) if the env/ide overrides & EditorConfig inclusion is deemed undesirable to the project.

### DIFF
--- a/styx/core/styx-cpu-pcode-backend/styx-sla/build.rs
+++ b/styx/core/styx-cpu-pcode-backend/styx-sla/build.rs
@@ -4,7 +4,7 @@ use std::{
     env,
     fs::File,
     io::{BufWriter, Write},
-    path::PathBuf,
+    path::{Path, PathBuf},
 };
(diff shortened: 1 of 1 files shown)


YurBoiRene:
Nice find!

I don't think we need to add the editor config, vscode settings, or gitattributes for the line endings as those can be configured on users' systems. However, support the line ending normalization codepath in the sleigh patching. We've had Windows/WSL devs work on the project before but I guess they didn't have autocrlf enabled :/

Can you make this PR only contain ae745f1 and we should be good :)

YurBoiRene (approved):

PR #115, opened 2025-09-29
Ci image fixes

YurBoiRene:


lockbox:
testing out now

lockbox:
These fixes work for me, thank you 😄 

> Also broke up the steps so its easier to restart from a checkpoint but let me know if this shouldn't be this way.

Yeah, generally you want to reduce the number of `RUN` statements to decrease docker image size since each one adds another layer. though honestly i think this is fine 🤷

PR #116, opened 2025-09-30
fix: rust API docs redirect

KevOrr:
After running `just rust-docs`, the instructions indicate to open `./target/doc/index.html`, but when you do so, the meta refresh `<meta http-equiv="refresh" content="0; URL=styx/index.html"/>` instructs the browser to open `./target/doc/styx/index.html`, which doesn't exist. I believe the right file is `./target/doc/styx_emulator/index.html`.


YurBoiRene (approved):
Good find, thanks!

PR #118, opened 2025-10-05
fix: workspace builds on non linux systems

lockbox:
i missed getting this into the fixes branch. apologies. now full workspace builds will work (honestly no clue how i missed this).

Before this i just gated all of the cow stuff with the global `#![cfg(target_os = "linux")], which is insufficient as then there will be no `fn main()` in the `main.rs` and the benchmark for the CoW code, both of which are included in `cargo build --workspace`... 🫠 Should be fixed now


lockbox:
also added 5c2ec21 while I was at it

PR #119, opened 2025-10-06
chore: add error contexts to loader operations

KevOrr:
Add some `anyhow::Context`s to some loader operations

### DIFF
--- a/styx/core/styx-processor/src/processor/builder.rs
+++ b/styx/core/styx-processor/src/processor/builder.rs
@@ -36,9 +36,11 @@ enum TargetProgramSource<'a> {
 }
 
 impl<'a> TargetProgramSource<'a> {
-    pub fn bytes(self) -> Result<Cow<'a, [u8]>, std::io::Error> {
+    pub fn bytes(self) -> Result<Cow<'a, [u8]>, styx_errors::UnknownError> {
         match self {
-            TargetProgramSource::File(file_name) => std::fs::read(file_name).map(Cow::Owned),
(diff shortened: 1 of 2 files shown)


styx/core/styx-processor/src/processor/builder.rs
```diff
@@ -36,9 +36,11 @@ enum TargetProgramSource<'a> {
 }
 
 impl<'a> TargetProgramSource<'a> {
-    pub fn bytes(self) -> Result<Cow<'a, [u8]>, std::io::Error> {
+    pub fn bytes(self) -> Result<Cow<'a, [u8]>, styx_errors::UnknownError> {
```
KevOrr, line 39: This is the only one I'm not really sure about. I modified the behavior inside `TargetProgramSource::bytes()` instead of where `bytes()` is called because `bytes()` moves `self`, so you can't match on `self` after `bytes()` returns an error (I think? I'm a Rust newb so correct me if I'm wrong). But because that means we're attaching the context within `bytes()`, we have to change the Result error type to `UnknownError`/`anyhow::Error` right?

YurBoiRene: Correct! I think this is a good change. The tradeoff with this change is that the caller loses the info that the only possible error is an io error, but that's not important for this function I don't think.

Really, I think this method should take `&self` since Cow is a cheap clone and `std::fs::read()` should be able to take a `&str`. But we don't need to do this change here.

YurBoiRene (approved):

Issue #99, opened 2025-09-05
styx macro utility: StyxManifest


YurBoiRene:
**Is your feature request related to a problem? Please describe.**
When developing the derive macro for `ProcessorConfig`, I realized that I had no way to reference the `styx_processor` crate. In the crate the reference would be  `$crate::module::ProcessorConfig`, in core crates it is `::styx_processor::module::ProcessorConfig`, in styx crates it would be `::styx_core::processor::module::ProcessorConfig`, and outside styx it would be `::styx::processor::module::ProcessorConfig`. Without making `ProcessorConfig` a dependency of the styx-macro crate, I don't know how to resolve these.

**Describe the solution you'd like**
I'd like to use `#[derive(ProcessorConfig)]` from any styx or external crate.

**Additional context**
I believe we could use Bevy's [`BevyManifest`](https://docs.rs/bevy_macro_utils/latest/bevy_macro_utils/struct.BevyManifest.html) to do this.

[`BevyManifest::maybe_get_path()`](https://docs.rs/bevy_macro_utils/latest/src/bevy_macro_utils/bevy_manifest.rs.html#83-110) is most of the way there ([example usage](https://docs.rs/bevy_ecs_macros/0.16.1/src/bevy_ecs_macros/states.rs.html#7-35)).

PR #121, opened 2025-10-07
feat: processor config

YurBoiRene:
1. Moved the styx_loader crate into styx_processor to avoid dependency issues arising from config types.
2. Added Proccesor config
3. Added StyxManifest #99 

The idea here is to allow arbitrary configuration types to be passed to the processor builder. Components of the processor (plugins, executor, processor impl) can pick these up and use them for configuration, error if a desired configuration is not found, or simply ignore them.

The untyped storage of the config allows for external crates to define and use their own configuration while allowing styx-processor to define a set of "standard" configuration values, think Arch, ExecutionBehavior, abstract peripheral config, etc. Currently these are typed on the ProcessorBuilder but I could see them being moved to this config paradigm as it is more flexible.

I think it is similar to the dependency injection design where components only use the configuration that they actually need. So if no component needs an Arch configured (for a raw processor or whatever) then theoretically no Arch config needs to be defined.


styx/core/styx-macros/src/lib.rs
```diff
@@ -952,3 +954,43 @@ pub fn enum_mirror(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+/// ```ignore
+/// use styx_macros::ProcessorConfig;
+///
+/// #[derive(ProcessorConfig)]
+/// pub struct MyPluginConfig {
+///     pub sample_rate: u32,
+/// }
+/// ```
+///
+/// expands to:
```
yuv418: nit: would the original structure definition still be preserved in addition to the trait impl?

styx/core/styx-macros/src/styx_manifest.rs
```diff
@@ -0,0 +1,203 @@
...
+    /// Calls `f` with a global shared instance of [`StyxManifest`].
+    pub fn shared<R>(f: impl FnOnce(&StyxManifest) -> R) -> R {
+        static MANIFESTS: RwLock<BTreeMap<PathBuf, StyxManifest>> = RwLock::new(BTreeMap::new());
+        let manifest_path = Self::get_manifest_path();
+        let modified_time = Self::get_manifest_modified_time(&manifest_path)
+            .expect("The Cargo.toml should have a modified time");
+
+        let manifests = MANIFESTS.read().unwrap_or_else(PoisonError::into_inner);
+        if let Some(manifest) = manifests.get(&manifest_path) {
+            if manifest.modified_time == modified_time {
```
yuv418, line 41: If this is for caching purposes, maybe it could be good to add a comment for that

styx/core/styx-macros/src/styx_manifest.rs
```diff
@@ -0,0 +1,203 @@
...
+/// code, since the correct path depends on which layer of the
+/// workspace the calling crate lives in.
+#[derive(Debug)]
+pub struct StyxManifest {
+    manifest: Document<Box<str>>,
+    modified_time: SystemTime,
+}
+
+impl StyxManifest {
+    /// Calls `f` with a global shared instance of [`StyxManifest`].
```
yuv418: Maybe worth adding a comment it's for the manifest in the current package (or however it's obtained from the MANIFESTS map)

styx/core/styx-processor/src/executor/default.rs
```diff
@@ -27,4 +29,12 @@ impl ExecutorImpl for DefaultExecutor {
         proc.cpu
             .execute(&mut proc.mmu, &mut proc.event_controller, insns)
     }
+
+    fn init(&mut self, proc: &mut crate::processor::BuildingProcessor) -> Result<(), UnknownError> {
+        self.stride_length = proc
+            .config
+            .get_or_default::<crate::processor::core_configs::ConfigRequestedStrideLength>()
```
yuv418: big nit: this import feels a bit long

yuv418: Nonetheless, I like how this is used here

styx/core/styx-processor/src/executor/default.rs
```diff
@@ -6,17 +6,19 @@ use crate::{core::ProcessorCore, cpu::ExecutionReport};
...
 use super::ExecutorImpl;
 
-#[derive(Default, Debug)]
 /// A sane default.
 ///
 /// Executes 1000 instructions per stride, handling events at the end of each stride.
 ///
 /// Notably implements [ExecutorImpl].
-pub struct DefaultExecutor;
+#[derive(Default, Debug)]
```
yuv418, line 21: ~~Is it worth setting the default impl to do 1000 for the stride?~~

I think the lifecycle of when init is called and what it does could be noted here (and in the other place where I mentioned it), since I was thinking that the DefaultExecutor would run with a stride of zero (but then I realized that `init` gets called before all of this)

styx/core/styx-processor/src/executor/test.rs
```diff
@@ -235,7 +235,11 @@ pub fn normal_begin_executor(
 /// Test the event cycle of the [`DefaultExecutor`]
 #[test]
 fn test_default() {
-    let executor = DefaultExecutor;
+    // This would normally be set via config but we set manually for test
+    // to avoid mocking all of BuildingProcessor.
+    let executor = DefaultExecutor {
+        stride_length: 1000,
```
yuv418, line 241: could this be easier by overriding the `Default` implementation in `DefaultExecutor`?

styx/core/styx-processor/src/loader/loaders/elf.rs
```diff
@@ -37,7 +40,6 @@ impl From<ElfMachine> for Arch {
             EM_RISCV => Arch::Riscv,
             EM_SH => Arch::SuperH,
             EM_AVR | EM_AVR32 => Arch::Avr,
-            EM_QDSP6 => Arch::Hexagon,
```
yuv418, line 41: I think this should still be there, since it'll break loading hexagon firmware without

styx/core/styx-processor/src/loader/loaders/elf.rs
```diff
@@ -46,29 +48,17 @@ impl From<ElfMachine> for Arch {
     }
 }
 
-/// Used in [`ElfLoaderConfig`] to dertermin which member to use for styx memory region address.
-#[derive(Debug)]
-pub enum SegmentAddr {
-    /// Use p_vaddr for segment address.
```
yuv418, line 53: I think this should be restored too, unless there's a reason it was removed

styx/core/styx-processor/src/loader/loaders/elf.rs
```diff
@@ -79,28 +69,14 @@ impl Default for ElfLoaderConfig {
...
-///
+/// # Available Hint's
 /// - if provided, an `endian` hint of type [`styx_cpu_type::ArchEndian`] can be provided,
 ///   it will not override the header, but will warn you if there is a disagreement
 /// - if provided, an `arch` hint of type [`styx_cpu_type::Arch`] can be provided,
 ///   it will not override the header, but will warn you if there is a disagreement
 /// - if provided, a `pc` hint of type [`u64`] can be provided, this *will* override
 ///   the header.
 ///
-/// ## p_paddr vs p_vaddr
```
yuv418, line 92: I think restore here as well

styx/core/styx-processor/src/loader/loaders/parameterized.rs
```diff
@@ -36,13 +37,6 @@ pub struct LoadFileElf {
     pub file: String,
 }
 
-/// This record structure specifies the parameters for a load of an Intel HEX file.
-#[derive(Deserialize, PartialEq, Debug, Clone)]
```
yuv418, line 40: Is this just a rebasing thing?

styx/core/styx-processor/src/processor/config/core_configs.rs
```diff
@@ -0,0 +1,42 @@
...
+/// [`ExecutorImpl`](crate::executor::ExecutorImpl) implementations use
+/// this as a hint when scheduling execution strides. The actual stride
+/// may differ for example in the gdb executor where the condition of
+/// 1 or more watchpoints will change the stride length to 1.
+#[derive(Debug, Clone, Copy, ProcessorConfig)]
+pub struct ConfigRequestedStrideLength {
+    pub preferred_stride_length: u64,
+}
+
+impl Default for ConfigRequestedStrideLength {
```
yuv418, line 36: Ah, I see why the "Default" impl didn't have this

util/xtask/src/hakari.rs
```diff
@@ -38,7 +38,7 @@ const HAKARI_GENERATE_FAIL_MSG: &str = r#"
     To update the installation please run the following commands and
     then commit the changes:
 
-    cargo xtask hakari --verify
```
yuv418, line 41: curious why this was changed

styx/core/styx-processor/Cargo.toml
```diff
@@ -39,7 +40,11 @@ tap = { workspace = true }
 nix = { workspace = true, features = ["signal"] }
 smallvec = { workspace = true }
 itertools = { workspace = true }
+goblin = { workspace = true }
```
yuv418, line 44: are these changes connected to the pr?

examples/raw-processor/src/main.rs
```diff
@@ -17,7 +17,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             ArchEndian::BigEndian,
         ))
         .with_loader(ParameterizedLoader::default())
-        .with_executor(DefaultExecutor)
+        .with_executor(DefaultExecutor::default())
```
yuv418, line 20: will the stride count be zero?

styx/core/styx-macros/src/styx_manifest.rs
```diff
@@ -0,0 +1,203 @@
...
+    #[test_case(STYX_TOML, "::styx_core")]
+    #[test_case(EXTERNAL_TOML, "::styx_emulator::processor")]
+    fn test_styx_processor_path(toml: &str, path: &str) {
+        let manifest =
+            Document::parse(toml.to_string().into_boxed_str()).expect("invalid manifest");
+
+        let styx_manifest = StyxManifest {
+            manifest,
+            // doesn't matter
+            modified_time: SystemTime::now(),
```
yuv418, line 201: Could it be worth testing the caching (if that is what this was for)?

styx/core/styx-processor/example-input/parameterized.yaml
```diff
@@ -11,9 +10,6 @@
     base: 0x10000
     # ELF file backing this region
     file: foo.elf
-- !FileIhex
```
yuv418, line 14: Is this intentional or the result of rebasing related things?

styx/core/styx-processor/src/executor/executor_impl.rs
```diff
@@ -145,4 +146,8 @@ pub trait ExecutorImpl: Send {
     fn get_stride_length(&self) -> u64 {
         1000
     }
+
+    fn init(&mut self, _proc: &mut BuildingProcessor) -> Result<(), UnknownError> {
```
yuv418: Does this need to be documented in the "executor loop" comment that's before this?

yuv418 (commented):
Overall looks good, maybe some docs changes/nitpicks in some places and also I think some things in main were removed (maybe a bad rebase)? 

I like this approach, it's clever and cool 😎

styx/core/styx-macros/src/styx_manifest.rs
yuv418: it == the manifest

styx/core/styx-processor/Cargo.toml
yuv418: is this related to styx loader? Maybe worth a comment?

styx/core/styx-processor/src/processor/builder.rs
```diff
@@ -228,15 +231,30 @@ impl<'a> ProcessorBuilder<'a> {
         self
     }
 
+    /// Overwrites a config parameter.
+    pub fn add_config<C: ProcessorConfig>(mut self, config: C) -> Self {
```
yuv418: I don't know where to put this, but I feel like it could be beneficial to add a documentation example of how to use this

styx/core/styx-processor/src/processor/builder.rs
```diff
@@ -228,15 +231,30 @@ impl<'a> ProcessorBuilder<'a> {
...
+        self.config.add_config(config);
+        self
+    }
+
+    /// Modify an existing config value (to e.g. add a value to a list) or modify the default.
+    pub fn modify_config_or_default<C: ProcessorConfig + Default>(
```
yuv418, line 297: I don't know where to put this, but I feel like it could be beneficial to add a documentation example of how to use this

styx/core/styx-macros/src/lib.rs
YurBoiRene: Good catch

styx/core/styx-processor/src/loader/loaders/elf.rs
YurBoiRene: Good catch here, I agree I think this was a rebase mess up caused because I moved the file :/

styx/core/styx-processor/Cargo.toml
YurBoiRene: Yeah, these are all deps for the loaders moved into styx-processor, added comments to these deps too.

styx/core/styx-macros/src/styx_manifest.rs
YurBoiRene: Added test

styx/core/styx-processor/src/executor/executor_impl.rs
YurBoiRene: Added !

styx/core/styx-loader-mock/src/lib.rs
yuv418, line 1: Out of curiosity, why was this removed (as opposed to being moved or updated to use the moved styx_loader)

styx/core/styx-processor/src/loader/mod.rs
```diff
@@ -107,25 +107,10 @@ pub trait Loader: std::fmt::Debug + Send {
         &self,
         data: Cow<[u8]>,
         hints: LoaderHints,
+        config: &Config,
     ) -> Result<MemoryLoaderDesc, StyxLoaderError>;
 }
 
-#[cfg(test)]
-mock! {
```
yuv418, line 114: Same question here about why this was removed

justfile
```diff
@@ -121,6 +121,11 @@ lint-docs:
 
     RUSTDOCFLAGS='--deny warnings' cargo doc --no-deps --all-features --workspace --bins --lib --examples --keep-going --document-private-items
 
+# Lint docs for a single package.
+lint-doc PACKAGE:
```
yuv418, line 125: Where is this used?

styx/core/styx-processor/src/processor/config/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+    ///
+    /// ```
+    /// use styx_processor::processor::{Config, ProcessorConfig};
+    ///
+    /// struct MyConfig { stride: u64 }
+    /// impl ProcessorConfig for MyConfig {}
+    ///
+    /// let config = Config::default();
+    ///
+    /// // Not registered, so `get` returns `None`.
```
yuv418, line 120: What does registered look like? Does registering a type involve calling `add_config`?

styx/core/styx-processor/src/processor/config/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+/// // `get_or_default` falls back to `Default::default()`.
+/// assert_eq!(config.get_or_default::<MyConfig>().stride, 0);
+/// ```
+#[derive(Default)]
+pub struct Config {
+    configs: HashMap<ConfigId, Box<dyn ProcessorConfig>>,
+}
+
+impl Config {
+    /// Inserts or overwrites the config value for type `C`.
```
yuv418: I think maybe use the "registered" verbiage here to tie things together (if this is the appropriate place to say "register a type" or whatnot)

yuv418 (changes requested):
Just one miniscule documentation fix and I'm curious why the mock stuff was removed, but looks very good otherwise 👍 💯

styx/core/styx-processor/src/processor/config/mod.rs
YurBoiRene: Yes, that is it. I will rename add_config to register_config.

justfile
YurBoiRene: I use it to lint a single package's docs that I am modifying. Useful because lint-docs doesn't stop on errors.

styx/core/styx-loader-mock/src/lib.rs
YurBoiRene: Correct me if I am wrong but I don't think it was used anywhere.

justfile
YurBoiRene: e.g. during my dev process

styx/core/styx-loader-mock/src/lib.rs
yuv418: I think you're right. I got confused by the `#[cfg(test)]` attribute used somewhere but I don't think anything is actually used after reading the `mockall` documentation.

yuv418 (approved):
Looks good 😎

PR #123, opened 2025-10-09
Add ihex loader

lockbox:
adds an ihex loader and another betaflight test firmware. feel free to tell me to nuke the new firmware and stick with the old one

### DIFF
--- a/styx/core/styx-loader/tests/test_ihex_elf_comparison.rs
+++ b/styx/core/styx-loader/tests/test_ihex_elf_comparison.rs
@@ -133,8 +133,8 @@ fn test_elf_ihex_range_consistency() {
     let elf_regions = elf_desc.take_memory_regions();
     let ihex_regions = ihex_desc.take_memory_regions();
 
-    let elf_bank = MemoryBank::from_regions_vec(elf_regions);
-    let ihex_bank = MemoryBank::from_regions_vec(ihex_regions);
+    let elf_bank = MemoryBank::from_regions_iter(elf_regions).unwrap();
+    let ihex_bank = MemoryBank::from_regions_iter(ihex_regions).unwrap();
 
     let elf_range = elf_bank.get_range().unwrap();
     let ihex_range = ihex_bank.get_range().unwrap();
--- a/styx/core/styx-memory/src/lib.rs
+++ b/styx/core/styx-memory/src/lib.rs
@@ -12,7 +12,33 @@ use zstd::{decode_all, encode_all};
 
 /// Container struct representing multiple [`MemoryRegion`]'s.
 /// Keeps a running tally of the min and max address while adding
-/// new [`MemoryRegion`]'s to the [`MemoryBank`] for faster queries
+/// new [`MemoryRegion`]'s to the [`MemoryBank`] for faster queries.
+///
+/// # Usage
+/// Construct a default and add as needed, or construct a [`MemoryBank`]
+/// from a collection of [`MemoryRegion`]'s you already have.
+///
+/// ```rust
+/// use styx_memory::{MemoryBank, MemoryRegion, MemoryPermissions as Perms};
+///
+/// // Method 1: Default construction and add regions one by one
+/// let bank = MemoryBank::default();
+/// bank.add_region(MemoryRegion::new(0x1000, 0x1000, Perms::READ | Perms::WRITE).unwrap()).unwrap();
+/// bank.add_region(MemoryRegion::new(0x3000, 0x1000, Perms::READ | Perms::EXEC).unwrap()).unwrap();
+///
+/// assert!(bank.contains_address(0x1000));
+/// assert!(bank.contains_address(0x3500));
+///
+/// // Method 2: Construct from a Vec of MemoryRegion using TryFrom
+/// let regions = vec![
+///     MemoryRegion::new(0x1000, 0x1000, Perms::READ | Perms::WRITE).unwrap(),
+///     MemoryRegion::new(0x3000, 0x1000, Perms::READ | Perms::EXEC).unwrap(),
+/// ];
+/// let bank2 = MemoryBank::try_from(regions).unwrap();
+///
+/// assert!(bank2.contains_address(0x1000));
+/// assert!(bank2.contains_address(0x3500));
+/// ```
 #[derive(Debug)]
 pub struct MemoryBank {
     /// Inner collection of [`MemoryRegion`]'s inside the
@@ -41,10 +67,53 @@ impl Default for MemoryBank {
     }
 }
 
-unsafe impl Send for MemoryBank {}
-unsafe impl Sync for MemoryBank {}
+impl TryFrom<Vec<MemoryRegion>> for MemoryBank {
+    type Error = StyxMemoryError;
+
+    fn try_from(value: Vec<MemoryRegion>) -> Result<MemoryBank, StyxMemoryError> {
+        Self::from_regions_iter(value)
+    }
+}
 
 impl MemoryBank {
+    /// Gets the representative [`MemoryRange`] for the [`MemoryBank`]
+    pub fn get_range(&self) -> Result<MemoryRange, StyxMemoryError> {
+        Ok(MemoryRange::new(self.min_address()?, self.max_address()?))
+    }
+
+    /// Constructs a [`MemoryBank`] from an [`Iterator`] like container
+    /// of [`MemoryRegion`]'s. The regions are added sequentially, and
+    /// if any region fails to be added (e.g., due to overlapping with
+    /// an existing region), an error is returned.
+    ///
+    /// ```rust
+    /// use styx_memory::{MemoryBank, MemoryRegion, MemoryPermissions as Perms};
+    ///
+    /// let regions = vec![
+    ///     MemoryRegion::new(0x1000, 0x1000, Perms::READ | Perms::WRITE).unwrap(),
+    ///     MemoryRegion::new(0x3000, 0x1000, Perms::READ | Perms::EXEC).unwrap(),
+    /// ];
+    ///
+    /// let bank = MemoryBank::from_regions_iter(regions.clone()).unwrap();
+    /// // equivalently:
+    /// let bank2 = MemoryBank::try_from(regions).unwrap();
+    /// assert!(bank.contains_address(0x1000));
+    /// assert!(bank.contains_address(0x3000));
+    /// # assert!(bank2.contains_address(0x1000));
+    /// # assert!(bank2.contains_address(0x3000));
+    /// ```
+    pub fn from_regions_iter(
+        regions: impl IntoIterator<Item = MemoryRegion>,
+    ) -> Result<Self, StyxMemoryError> {
+        let bank = MemoryBank::default();
+
+        for region in regions.into_iter() {
+            bank.add_region(region)?;
+        }
+
+        Ok(bank)
+    }
+
     /// Get's the current minimum address represented in the
     /// [`MemoryBank`], if there are no [`MemoryRegion`]'s in the
     /// [`MemoryBank`] then this methods returns an error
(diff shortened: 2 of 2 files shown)


styx/core/styx-loader/tests/test_ihex_elf_comparison.rs
```diff
@@ -0,0 +1,299 @@
...
+//! consistent memory mappings when loading the same binary in different formats.
+
+use std::borrow::Cow;
+use std::collections::HashMap;
+use std::fs;
+use std::path::PathBuf;
+use styx_loader::{ElfLoader, IhexLoader, Loader};
+
+/// Helper function to get the path to test data
+fn test_data_path(filename: &str) -> PathBuf {
```
YurBoiRene, line 18: We already have this, `styx_util::resolve_test_bin()`

YurBoiRene: Ah I see it's in styx-loader/test-data, not data/test-binaries

styx/core/styx-loader/test-data/hello-world/README.md
```diff
@@ -0,0 +1,146 @@
...
+```bash
+# Run all Intel HEX tests
+cargo test test_ihex
+
+# Run the comparison tests specifically
+cargo test --test test_ihex_elf_comparison
+
+# Run with output to see the verification details
+cargo test test_elf_ihex_memory_consistency -- --nocapture
+```
```
YurBoiRene: These are weird, I would make these use `cargo nextest` and also give the commands in the form that you would run from the styx root, i.e. `cargo nextest --package styx-loader ....`

styx/core/styx-loader/test-data/hello-world/README.md
```diff
@@ -0,0 +1,146 @@
...
+- Data records (Type 00): Contains actual program data
+- End of File (Type 01): Marks the end of the file
+- Extended Linear Address (Type 04): Sets upper 16 bits of address
+- Start Linear Address (Type 05): Specifies the entry point
+
+## References
+
+- [Intel HEX Format Specification](https://en.wikipedia.org/wiki/Intel_HEX)
+- [ELF Format Documentation](https://refspecs.linuxfoundation.org/elf/elf.pdf)
+- [ARM Cortex-M Memory Map](https://developer.arm.com/documentation/)
```
YurBoiRene: Weird link to all of arm documentation

data/test-binaries/arm/stm32f405/bin/betaflight_4.5.1_STM32F405.hex
YurBoiRene, line 1: I would use the old one, I don't know the reason why would we want this one of the 4.5.0.

YurBoiRene: The tests pass with the old one as well

styx/core/styx-loader/tests/test_ihex.rs
```diff
@@ -0,0 +1,90 @@
...
+    let registers = desc.take_registers();
+
+    // The file has a Start Linear Address record (type 05) which should set the PC
+    if !registers.is_empty() {
+        println!("Loaded {} register(s):", registers.len());
+        for (reg, value) in &registers {
+            println!("  {reg:?} = 0x{value:08X}");
+        }
+    }
+}
```
YurBoiRene, line 88: This should assert pc is present and set to the correct value. At least that's what is suggested by the comments

styx/core/styx-loader/test-data/hello-world/Makefile
```diff
@@ -0,0 +1,91 @@
...
+          -Wl,-Map=hello.map \
+          -Wl,--gc-sections
+
+# Target files
+TARGET = hello
+ELF_FILE = $(TARGET).elf
+HEX_FILE = $(TARGET).hex
+BIN_FILE = $(TARGET).bin
+MAP_FILE = $(TARGET).map
+LST_FILE = $(TARGET).lst
```
YurBoiRene, line 35: I think you can remove the map and lst as they aren't used for tests. Maybe the bin as well since I don't think it's loaded either

styx/core/styx-loader/test-data/hello-world/README.md
```diff
@@ -0,0 +1,146 @@
...
+This layout is typical for STM32 microcontrollers and provides a realistic (and simple) test case.
+
+## Building the Test Binaries
+
+### Prerequisites
+
+You need the ARM embedded toolchain installed:
+- **macOS**: `brew install arm-none-eabi-gcc`
+- **Ubuntu/Debian**: `apt-get install gcc-arm-none-eabi`
+- **Arch Linux**: `pacman -S arm-none-eabi-gcc`
```
YurBoiRene: We have a dockerfile for the arm toolchain, this should instruct users to use that.

styx/core/styx-loader/test-data/hello-world/Makefile
```diff
@@ -0,0 +1,91 @@
+# SPDX-License-Identifier: BSD-2-Clause
+# Makefile for building hello world test binary in both ELF and Intel hex formats
+
+# Cross-compilation toolchain for ARM
```
YurBoiRene, line 4: Possibly  add a target to build with docker

styx/core/styx-loader/test-data/hello-world/hello.c
```diff
@@ -0,0 +1,38 @@
...
+const unsigned int magic_number __attribute__((section(".rodata"))) = 0xDEADBEEF;
+const unsigned int version_info __attribute__((section(".rodata"))) = 0x01020304;
+
+// Add a data section with some values
+unsigned int data_values[] __attribute__((section(".data"))) = {
+    0x11111111,
+    0x22222222,
+    0x33333333,
+    0x44444444
+};
```
YurBoiRene: This part is weird... while the rodata is technically a different section in the ELF, it has the same attributes as the executable code it is put into the same segement, which doesn't test any extra code in the hex/elf loaders.

The `.data` section would have a separate section with RWX perms but it does not show up (via `readelf --segments`), presumably because they aren't used. Since they don't add a program segment, they don't test any additional code in the loaders.

I would like this see this test have multiple sections, some continuous with previous sections and some disjoint.

styx/core/styx-loader/tests/test_ihex_elf_comparison.rs
lockbox: yeah, it didnt fit with the other test data we have since its loader specific and super minimal. do you want it moved?

data/test-binaries/arm/stm32f405/bin/betaflight_4.5.1_STM32F405.hex
lockbox: yeah good, they should. the old/new version was more i didnt realize till after i commited and pushed that we had the 4.5.0 in here. i should have a fun "machine" for one soon 😎

styx/core/styx-loader/test-data/hello-world/Makefile
lockbox: ok will do 👍 here i wasnt sure which things to keep to have provenance for "this fails why and from where" so i just added all the artifacts to make it easier to debug if a test went wrong etc.

styx/core/styx-loader/tests/test_ihex_elf_comparison.rs
YurBoiRene: No it's good there :+1:

styx/core/styx-loader/test-data/hello-world/Makefile
lockbox: Going to leave them being built in case someone wants to debug something later then they wont have to add that back in, but going to not commit them (they are gitignored by default anyways)

styx/core/styx-loader/test-data/hello-world/hello.c
lockbox: ihex doesn't do permissions so thats not really in scope here, i trimmed down on a lot of stuff and now i hate both objcopy and linkers more than i did this morning. this should be a lot tidier now.

styx/core/styx-loader/tests/test_ihex_elf_comparison.rs
```diff
@@ -0,0 +1,259 @@
...
+    // The gnu toolchain likes to change the number of
+    // emitted segments / sections etc. based on size or region, overall size
+    // of segment, and sizes of sections for both the ELF and Intel HEX.
+    //
+    // The loaders should have the same number of regions
+    // assert_eq!(
+    //     elf_regions.len(),
+    //     ihex_regions.len(),
+    //     "ELF and Intel HEX loaders produced different number of memory regions"
+    // );
```
YurBoiRene, line 71: Is the takeaway from this comment that the generated segments of Ihex and ELF will be different and future developers shouldn't try to make them have the same segments?

YurBoiRene: Additionally, it looks like ihex will always have more regions than the elf equivalent

styx/core/styx-memory/src/lib.rs
```diff
@@ -44,7 +44,43 @@ impl Default for MemoryBank {
...
+            }
+        });
+
+        Self {
+            regions: Arc::new(RwLock::new(regions)),
+            min_address: min_address.into(),
+            max_address: max_address.into(),
+        }
+    }
+
```
YurBoiRene: This should iterate over the regions and use `add_region` one by one in order to catch overlapping regions. Otherwise, good addition to the MemoryBank api :).

You could also change to take `impl IntoIterator<Item = MemoryRegion>` for funsies

YurBoiRene (changes requested):

YurBoiRene:
The additional sections are not present in the resulting ELF as additional sections, they are combined into one PT_LOAD segment. Is this intended?

lockbox:
answering the section / segment scheduling questions here since its pretty chaotic / implementation defined as to what toolchain places what code where. The main thing i was trying to test was whether the same content ends up at the same location.

- depending on the backing segment in the linker being in `RAM` or `ROM` the ihex would include / omit it from the content. 
- depending on the `PT_FILL` options, the elf would instead use that specified fill `u8` instead of `0x00`, but that also would not specifically dictate what the `ihex` would do.
- depending on the section names and section location the elf location scheduling would put sections and segments physically next to each other, but mapped in the same location, physically not next to each other and mapped into the same location etc. etc. in all the permutations. 
- objcopy behavior is not configurable, the linker and rest of the toolchain are, though.

The issue i ran into was that it quickly became a test of "what linker script, gcc, and objcopy options will result in something that looks the same" instead of "do our ihex loader and elf loader do the thing".

> Is the takeaway from this comment that the generated segments of Ihex and ELF will be different and future developers shouldn't try to make them have the same segments?

Yes, depending on the segment and section options the toolchains (not just the linkers) seem to place things in different locations based on the linker section options and segments they are destined for. `objcopy` exporting `ihex` wont include things that are backed by RAM. (would need extra invocation of explicit section name maybe? idk, spent way too long trying to get things to line up). the main problem IMO is that everything except objcopy is configurable.....and that is kind of an important piece to be able to control so that we dont have to manually create ihex files. which at some point i probably should have and removed that variable.

> Additionally, it looks like ihex will always have more regions than the elf equivalent

Not necessarily, that is true in this case (where `gnu-arm-non-eabi-objcopy` from ubuntu-22.04 was used to convert from an elf to an ihex). see above complaints about toolchains 🫠 

### What do next

These tests currently show that all the content shows up properly from the same source file and gets loaded by the `styx-loaders` (ignoring the fact that the elf file explicitly sets a bunch of memory to `0x00`) in a manner that is equivalent for a real use case. at some point for a more development oriented focus it might make sense to address some of the issues i mentioned above, but that is (imo) out of scope for this project, and this ticket.

TLDR; it loads ihex as expected 👍

styx/core/styx-memory/src/lib.rs
lockbox: cool, i initially decided against that probably for some reason that didnt make sense 😆 ill also add a `From<T>` for the same unless you think that's too much magic

YurBoiRene: I think it would be `TryFrom<T>` since it would error on overlapping regions but yeah sounds useful

YurBoiRene:
> answering the section / segment scheduling questions here since its pretty chaotic / implementation defined as to what toolchain places what code where. ...

Sounds good to me. I apologize, I don't think I had a grasp of the differences between elf segments and ihex data "segments" and didn't mean to put unnecessary and practically unabtainable constraints on the testing of the loader. On the bright side this should be good documentation for someone in the future :)

After the styx-memory `from_regions()` change this should be gtg.

PR #124, opened 2025-10-14
Elf loader choose to use p_vaddr or p_paddr

YurBoiRene:
In the future this would probably be a Config option.

Also added powerpc compile container to `util/docker`.

### DIFF
--- a/styx/core/styx-loader/src/loaders/elf.rs
+++ b/styx/core/styx-loader/src/loaders/elf.rs
@@ -45,17 +45,29 @@ impl From<ElfMachine> for Arch {
     }
 }
 
+/// Used in [`ElfLoaderConfig`] to dertermin which member to use for styx memory region address.
+#[derive(Debug)]
+pub enum SegmentAddr {
+    /// Use p_vaddr for segment address.
+    Vaddr,
+    /// Use p_paddr for segment address.
+    Paddr,
+}
+
 /// Configure the [`ElfLoader`].
 #[derive(Debug)]
 pub struct ElfLoaderConfig {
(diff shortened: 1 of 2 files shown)


styx/core/styx-loader/src/loaders/elf.rs
```diff
@@ -45,17 +45,33 @@ impl From<ElfMachine> for Arch {
...
     pub warn_no_loadable_segments: bool,
+    /// Use this `p_paddr` or `p_vaddr` for region start address. Defaults to `p_vaddr`.
+    pub segment_addr_preference: SegmentAddr,
 }
 
 impl Default for ElfLoaderConfig {
     fn default() -> Self {
         Self {
             warn_no_loadable_segments: true,
+            segment_addr_preference: SegmentAddr::Vaddr,
```
yuv418, line 70: maybe move the old comment about loading firmware files ("note in general we load firmware files so p_vaddr/p_paddr should be set to the same value".....), might be a good hint to someone tackling this for the first time

miniflux.container
```diff
@@ -0,0 +1,7 @@
+# miniflux.container
```
lockbox: what is the miniflux file here for?

YurBoiRene: LOL whoopsies, removed

styx/core/styx-loader/src/loaders/elf.rs
YurBoiRene: I originally moved that comment to `SegementAddr` but you reminded me that it should be visible to someone using for the first time so I made a section in the `ElfLoader` struct doc.

yuv418 (approved):
looks good (:

PR #37, opened 2025-08-18
feat: add hexagon test binaries

YurBoiRene:
closes #35

PR #129, opened 2025-10-25
Hexagon ISA fixes, some initial system instructions

yuv418:
This PR is a "big tent" that does some fixes, additions, and the like for Hexagon.

It has
- Adding a bunch of system registers + updates everywhere for that (this is needed for `cswi`/`ciad` that use the newly-added IPEND and IAD regs)
- The `cswi` instruction + a test. 
- The `ciad` instruction + a test
- The `trap0` instruction + a test
- Fixes for `call` instruction not setting the link register + 4 test cases 
- Bitfields plucked from QEMU for some system registers
- Some modifications to enable initial functionality in the tests merged in #37. Note that these don't work quite yet since those tests seem to depend on the Linux syscall ABI. 
- A random fix I had to do to get LLDB to work at all

Might have to do:
- Rewrite a comment on how I obtained LLDB register mappings in first-person plural.
- Verify the LLDB fixes didn't break GDB functionality for other targets.

Note that `cswi` and `ciad` are simple, but not implemented in P-code. I think it's possible we will need to do integration that isn't available in P-code with interrupt controllers, so I have left their implementations with the current callother-based implementations for now. 

I can also split the PR into smaller chunks if it helps/feels more focused. TLB coming after this, its implementation seems to work and just have to polish that up  :+1:

### DIFF
--- a/styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/execution_helper.rs
+++ b/styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/execution_helper.rs
@@ -121,9 +121,14 @@ impl DefaultHexagonExecutionHelper {
     }
 
     /// Hook at the beginning of a packet to determine
-    /// what the next packet start is. This is important
-    /// for if a call is in the packet, as the packet
-    /// must set the link register to the start of the next packet.
+    /// what the next packet start is.
+    ///
+    /// Hexagon call instructions must must set the link
+    /// register to the start of the next packet.
+    ///
+    /// To do this, we set the pkt_next context option to
+    /// the value of the next packet, which Sleigh in turn
+    /// uses to set the link register during a call.
     ///
     /// Section 11.4 "Call subroutine" details this.
     fn detect_pkt_next(
--- a/styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/system/interrupt.rs
+++ b/styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/system/interrupt.rs
@@ -82,10 +82,10 @@ impl<T: CpuBackend> CallOtherCallback<T> for CswiHandler {
     /// seem to define the mask to clear the lower 16 bits of the pending interrupts register,
     /// even though the mask is defined in the manual 11.9.2 "Clear Pending Interrupts" to be
     /// 32 bits.
-    //
+    ///
     /// Other branches, such as bcain/tlb_obj, seem to define a separate register IPEND and IAD
     /// as 32-bit registers. These registers come towards the end of the defined list of registers.
-    //
+    ///
     /// In 11.9.2 "System control register transfer," however, the registers IPEND and IAD
     /// are defined separately as 32-bit values, but in the same place where both QEMU branches
     /// define IPENDAD and VID1, respectively. We have chosen to move IPEND and IAD to the end of
--- a/styx/core/styx-cpu-pcode-backend/tests/hexagon_tcg.rs
+++ b/styx/core/styx-cpu-pcode-backend/tests/hexagon_tcg.rs
@@ -87,7 +87,6 @@ use test_case::test_case;
 fn test_binutils_unittests(test: TestData) {
     styx_util::logging::init_logging();
 
-    // TODO change to hexagon
     let mut backend = HexagonPcodeBackend::new_engine(
         Arch::Hexagon,
         HexagonVariants::QDSP6V66,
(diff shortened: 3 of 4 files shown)


styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/system/interrupt.rs
```diff
@@ -15,6 +25,142 @@ pub struct InterruptGenericStub {
...
+#[derive(Debug)]
+pub struct CswiHandler;
+impl<T: CpuBackend> CallOtherCallback<T> for CswiHandler {
+    /// Implement "Clear pending interrupts."
+    ///
+    /// For some reason, the QEMU implementation (hex_interrupt.c and registers bitfields in branch hex-next)
+    /// seem to define the mask to clear the lower 16 bits of the pending interrupts register,
+    /// even though the mask is defined in the manual 11.9.2 "Clear Pending Interrupts" to be
+    /// 32 bits.
+    //
```
lockbox: please ensure the comments are all the same `///` level here

styx/core/styx-cpu-pcode-backend/tests/hexagon_tcg.rs
```diff
@@ -123,14 +129,14 @@ fn test_binutils_unittests(test: TestData) {
 
     // TODO change to hexagon register
```
lockbox: please remove these OBE todo's

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/execution_helper.rs
```diff
@@ -120,6 +120,40 @@ impl DefaultHexagonExecutionHelper {
         }
     }
 
+    /// Hook at the beginning of a packet to determine
+    /// what the next packet start is. This is important
+    /// for if a call is in the packet, as the packet
+    /// must set the link register to the start of the next packet.
+    ///
```
lockbox, line 132: could change to something like:

```rust
    /// Hook at the beginning of a packet to determine
    /// what the next packet start is. 
    /// 
    /// Hexagon Insn packets must set the link register
    /// to the start of the next packet, we do so here.
    ///
```

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/execution_helper.rs
```diff
@@ -120,6 +120,40 @@ impl DefaultHexagonExecutionHelper {
...
+            //
+            // For reference, see section 10.3, duplexes are slot 0 and 1.
+            // See also section 10.5, describes the different parse bits referenced below.
+            if matches!(
+                insn_value.parse(),
+                PktLoopParseBits::Duplex | PktLoopParseBits::EndOfPacket
+            ) {
+                // Each instruction is 4 bytes, and we want the instruction immediately following this.
+                let pc_offset_next = self.pc.unwrap() as u32 + (i as u32 + 1) * 4;
+                backend.update_context(
```
lockbox, line 152: more of an observation: gross global things like this are gnarly, and we should (if possible) in the future attempt to migrate this out of a global state struct into a more 'data oriented' approach that passes the state from one stage to the next to keep the 'iterate over insns in packet' stages to the minimum number of passes.

yuv418: Hmm, I think this warrants a separate issue/PR, because the global state for context stuff was something that was borne out of an old RFC. We can definitely pursue that further, though

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/system/interrupt.rs
yuv418: whoops! I'm actually surprised that the doc linting/whatnot didn't catch this.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/execution_helper.rs
yuv418: I updated the comment (with a little more detail), is it better now?

lockbox: Yeah it's good, for some reason i need perms to resolve this issue from my own review. great.

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/backend/execution_helper.rs
lockbox: Yeah def should be a new issue etc once we dial in all the features that we'd want. Treating is similar to the `ExecutionReport`, where the state of what's happening is a struct taht gets passed around to keep the state locked to the specific "pipeline" of execution instead of a global thing (also should help speed some things up, in theory). idk, just something to think about. 

Similar issues here, cannot resolve this issue.

PR #152, opened 2025-11-15
doc: add roadmap explainer in readme

YurBoiRene:


kudu-emilie:
@YurBoiRene modified the wording slightly. Doesn't look like I can set you as the approver since you're the original author, so give me a :+1: and I'll press the approve button

PR #155, opened 2025-12-03
feature: memory supports atomics

### LINKED ISSUE #146: Multi-processor Memory initial implmentation
The new Memory Structure should support reading/writing with only a immutable reference to `Memory`. Operations should be explicit about the nature of if they are atomic or not.

I envision each region has a continuous blob of `Atomicu64` to match the host's 64-bit platform. This would allow us to perform up to 64 bit atomic operations. 1-8 byte operations would consist of some sort of byte masking if necessary. >8 byte operations would not be atomic. This would be a limitation.

### Memory Region Mutability
What sort of region mutability do we want to offer at runtime? We could have append only regions using something like [boxcar](https://github.com/ibraheemdev/boxcar) or we could simply have no region mutability during runtime.

### Load-link/store-conditionals
Covered in this [article of QEMU's journey](https://lwn.net/Articles/697265/).

### Memory Caches
In the future, we may want to emulate memory caching on a per-core basis. This would mean each core does not get a `&Memory` but instead a `&VCPUMemory` which will hold a `&Memory` and can optionally implement caching and write buffers.

Should we include the per-vCPU memory structure in this initial implementation?

YurBoiRene:
Read some background in #146 

This completes the switch over of the memory backend (`MemoryRegion`) to use `AtomicU64` allowing for atomic operations and read/writes using immutable references to memory. I have not switched the uses of `&mut Mmu` to `&Mmu` but the methods on `Mmu` all take `&self` at this point.

Some other changes that ended up happening, but not directly related to atomic enabled memory.

- removed the flat memory spaces, they are still available from the public api, they are represented as a region store with a single region
- `AtomicWord` presents a byte addressable API for an `AtomicU64`
- Flattened the `memory` module in `styx-processor`
- `MemoryArchitecture` for getting Memory properties that will be different for different memory architectures

- compare and swap public api moved to #156 

Further work is needed on the memory side of things to specifically to support Hexagon's threading
- #156 
- #157

closes #146


incubation/styx-be-arm-processor/src/main.rs
```diff
@@ -209,16 +212,16 @@ mod test_machine {
         //these asserts verify that registers have been written to
         let value = machine.read_memory(0x2202_0000);
 
-        trace!("Value stored @ 0x2202_0000: {:x}", value);
+        println!("Value stored @ 0x2202_0000: {value:x}");
```
appar3ntly: Should these go back to `trace`, assume it was for test debugging?

styx/core/styx-cpu-unicorn-backend/src/lib.rs
```diff
@@ -682,20 +685,13 @@ impl UnicornBackend {
...
-                r.size(),
-                styx_to_unicorn_permissions(&r.perms).bits(),
-            )
-        });
+        let uc_regions = &self.unicorn_regions;
 
-        Ok(uc_regions.eq(mmu_regions))
+        let mmu_regions = mmu.regions();
+        let mmu_regions = get_unicorn_regions(mmu_regions)
+            .with_context(|| "current memory regions are not 4k aligned")?;
```
appar3ntly: Is the `with_context` necessary when non-4K alignment is already the `anyhow!`'ed Error that can pop out?

Edit: This applies throughout with this particular `with_context` pattern. It obviously doesn't hurt, it's more of a "could this hide/make confusing other errors if the regions have additional validation done at some point?" concern.

styx/core/styx-cpu-unicorn-backend/src/lib.rs
```diff
@@ -866,6 +855,73 @@ impl UnicornBackend {
...
+    fn try_from(region: &MemoryRegion) -> Result<Self, Self::Error> {
+        region
+            .expect_aligned(0x1000)
+            .map_err(|_| TryIntoUnicornError::UnalignedRegion)?;
+        let permissions = styx_to_unicorn_permissions(&region.perms());
+
+        let (data, size) = region
+            .as_raw_parts()
+            .ok_or(TryIntoUnicornError::EmptyRegion)?;
+        let size = NonZeroUsize::new(size)
```
appar3ntly: Not going to be a big stickler on this, it's just a "this seems like a lot of mixed up validation" nitpick, but this feels off somehow. If `as_raw_parts` can't return a zero sized ptr, this shouldn't be necessary.

Looking at it just a bit I'm not entirely sure if it could or not. It looks like the `data.0.is_empty` _could_ be decoupled from returning the `self.size as usize` as the size if something screwy happened during construction?

Without tracing everything, would it make sense to just add `self.size != 0` in the `as_raw_parts` `None` return path and untangle this at some point later?

styx/core/styx-cpu-unicorn-backend/src/lib.rs
```diff
@@ -1114,6 +1173,15 @@ mod tests {
...
         let mut machine = TestMachine::with_code("movw r1, #0x9999;ldr r4, [r1];");
+        let formatter = machine
+            .mmu
+            .format_regions()
+            .with_format(MemoryRegionFormat {
+                is_zeroed: true,
+                ..Default::default()
+            });
+
+        println!("{formatter:?}");
```
appar3ntly: More println leftovers?

styx/core/styx-processor/src/memory/atomic_word.rs
```diff
@@ -0,0 +1,305 @@
...
+    #[allow(unused)]
+    Success,
+    /// The swap was not performed.
+    #[allow(unused)]
+    Failure,
+}
+
+/// Copy a idx and size from an Atomic.
+///
+/// - idx and size MUST HAVE BEEN CHECKED FIRST.
```
appar3ntly, line 170: This makes sense, and I'm putting my nitpick hat on for structure stuff again.

Would it make sense and be more rust-y (rust-ic?) to incorporate this assumed naked read (and corresponding write) into the `impl AtomicWord`, and then give yourself:

* `read`
* `checked_read`
* `write`
* `checked_write`

Maybe corresponding `update`s as well I haven't looked at them.

styx/core/styx-processor/src/memory/atomic_word.rs
```diff
@@ -0,0 +1,305 @@
...
+
+    #[test_case(0, 0)]
+    #[test_case(0, 8)]
+    #[test_case(2, 2)]
+    #[test_case(4, 4)]
+    #[test_case(0, 4)]
+    #[test_case(1, 1)]
+    #[test_case(0, 1)]
+    #[test_case(6, 2)]
+    fn test_read_write_update(idx: usize, size: usize) {
```
appar3ntly, line 245: Also sort of nitpicky, don't let me assault the whole existing styx testing codebase. This is more me coming out of my internal styx silo (where I'm currently horribly uncovered on tests).

Should we split this up into read/write/update tests separately? I agree this is sufficient for "if this passes it surely works", it's just a lot of things to bundle into a unit test.

styx/core/styx-processor/src/memory/mmu.rs
```diff
@@ -48,11 +50,11 @@ pub enum MemoryType {
 /// [`DummyTlb`].
 ///
 /// For a processor ready default use [`Mmu::default_region_store()`].
-///
-/// Access physical memory using [`Mmu::memory()`].
 pub struct Mmu {
-    pub tlb: Box<dyn TlbImpl>,
-    pub memory: MemoryBackend,
+    // not need to be pub
```
appar3ntly: Nitpick remove comments as they won't mean anything to someone past this review I'd think.

styx/core/styx-processor/src/memory/physical.rs
```diff
@@ -0,0 +1,482 @@
...
+}
+
+/// Defines all of the valid address spaces
+#[derive(Deserialize, Debug, PartialEq)]
+pub enum Space {
+    Code,
+    Data,
+}
+
+/// This enumeration is specifies the memory permissions for an allocated memory region.
```
appar3ntly:
```suggestion
/// This enumeration specifies the memory permissions for an allocated memory region.
```

styx/core/styx-processor/src/memory/memory_region.rs
```diff
@@ -207,18 +235,44 @@ impl MemoryRegionData for MemoryRegionView<'_> {
...
     base: u64,
+    /// Size of region, in bytes.
     #[getset(get_copy = "pub")]
     size: u64,
     #[getset(get_copy = "pub")]
     perms: MemoryPermissions,
-    data: Arc<UnsafeCell<Vec<u8>>>,
-    // the actual size of the data being stored
-    effective_size: u64,
+    /// Zero sized if the permissions are empty
```
appar3ntly: Nitpicking on capitalization/punctuation for this and the following comments just because they're doc comments.

styx/core/styx-processor/src/memory/memory_region.rs
```diff
@@ -373,51 +413,70 @@ impl MemoryRegion {
...
+        // We just checked the permissions
+        self.read_data_unchecked(base, data)
+    }
+
+    /// Read and write the region using the memory api.
+    pub fn data(&self) -> RegionDataHelper {
+        RegionDataHelper(self)
+    }
+
+    /// Get the region's data at a pointer and size, probably for ffi.
```
appar3ntly:
```suggestion
    /// Get the region's data as a pointer and size, probably for ffi.
```

styx/core/styx-cpu-unicorn-backend/src/lib.rs
YurBoiRene: Good catch there. Removed the `with_context` here and another place since it will already be in the `anyhow` error :)

styx/core/styx-cpu-unicorn-backend/src/lib.rs
YurBoiRene: Yeah I think I understand what you're saying. To make the `as_raw_parts` signature more correct, we should make it return `NonZeroUsize`, that way we never have to do this check later

styx/core/styx-processor/src/memory/atomic_word.rs
YurBoiRene: Good idea but I don't think works. The functions here operate on `u64` as opposed to `AtomicU64`, that's the main difference, not the idx/size checking. I think I chose this abstraction for the `write_to_word()` method first which is more involved than the `read_from_word` counterpart.

This is important for the `update()` and `compare_and_swap()` because they will both require the same masking to translate between the "integerness" of `u64` and the `[u8]` that the APIs expect.

appar3ntly: Oh I see. I didn't appreciate the `.load` that has to happen to get the value out of the atomic so you get a `u64`.

That makes complete sense to unify that common `u64` <-> `[u8]` this way, ignore me.

styx/core/styx-processor/src/memory/atomic_word.rs
YurBoiRene: valid, I split these up into 3 separate tests, but call them from the same test so we can reuse the `idx + size` test case list

styx/core/styx-processor/src/memory/mmu.rs
YurBoiRene: Thanks :)

styx/core/styx-processor/src/memory/physical.rs
YurBoiRene: Wow I didn't know you could comment a suggest like that, good to know

PR #167, opened 2025-12-31
Gdb step instruction memory hook test and Virtual memory hook fix

YurBoiRene:


YurBoiRene:
@yuv418 Did you get a chance to try this out? Let me know if it fixed your problem.

yuv418:
On the IRQ front, it seems to work perfectly :+1:

styx/plugins/styx-gdbserver/src/options.rs
```diff
@@ -0,0 +1,41 @@
...
+    /// [`EventControllerImpl::next()`](styx_core::event_controller::EventControllerImpl::next()).
+    /// Setting this to [StepIRQs::Enabled] will enable these during stepping.
+    /// The frequency of ticking is set by [`GDBOptions::cpu_epoch`].
+    pub step_irqs: StepIRQs,
+    /// How many instructions to run before pausing execution to handle IRQs and GDB input.
+    ///
+    /// The default is `1024`.
+    /// The general tradeoff is larger epoch means faster execution but slower to activate IRQs
+    /// and tick peripherals.
+    pub cpu_epoch: u64,
```
yuv418, line 32: A question about this: are there now "two" locations where this "number of instructions before tick" is? Eg. here, for the debugger, but also there are values that are used in, for example, the `post_stride_processing` stuff. 

Not sure how but I'm thinking about a scenario where I would have to set this "Number of instructions before tick" value in two different places so when I run without the debugger and when I run with the debugger the same value is used for both.

styx/plugins/styx-gdbserver/src/target_impl.rs
```diff
@@ -279,132 +285,90 @@ where
...
+            self.exec_mode == ExecMode::Continue || self.options.step_irqs == StepIRQs::Enabled;
+        loop {
+            if self.step_cycles >= self.options.cpu_epoch {
+                // insert interrupt
+                _ = self
+                    .proc
+                    .event_controller
+                    .next(self.proc.cpu.as_mut(), &mut self.proc.mmu);
+
+                // assume 1 ns per instruction
```
yuv418: question because I'm not aware of it, but is this assumption made elsewhere in the code? Looks like ticking in the "normal way" when `post_stride_processing` is called uses the actual wall-clock time. I would also wonder if 1 ns / instruction can be tweaked elsewhere. Then again, I'm not sure if anything uses `delta.time` right now.

styx/plugins/styx-gdbserver/src/target_impl.rs
```diff
@@ -279,132 +285,90 @@ where
...
+                }
+            }
+
+            // if we're stepping then we are done after one step
+            if self.exec_mode == ExecMode::Step {
+                break RunEvent::Event(event_loop::Event::DoneStep);
             }
 
             // step until the range, instead of attempting to do
             // a bunch of fun math because variable length instruction (VLIW)
```
yuv418: I think vliw stands for "very long instruction word"

styx/integration-tests/src/gdb_core_integration_test_suite.rs
```diff
@@ -915,6 +916,125 @@ macro_rules! gdb_core_test_suite {
...
+
+                // add watchpoint
+                tracing::info!("adding watchpoint at 0x{WP_ONE:X}");
+                let watchpoint_id = harness.add_watchpoint(WP_ONE).unwrap();
+                let watchpoints: Vec<i64> = harness.list_watchpoints().unwrap();
+                assert!(watchpoints.contains(&watchpoint_id));
+
+                // hit watchpoint
+                harness.gdb_continue().unwrap();
+                std::thread::sleep(std::time::Duration::from_millis(500));
```
yuv418: could this make it flaky?

styx/integration-tests/src/gdb_core_integration_test_suite.rs
```diff
@@ -915,6 +916,125 @@ macro_rules! gdb_core_test_suite {
...
+                    $gdb_test_processor().add_hook(StyxHook::memory_write_virt(WP_ONE, hook));
+                let harness =
+                    ::styx_integration_tests::gdb_harness::GdbHarness::from_processor_builder::<
+                        GdbTestTargetDescriptionType,
+                    >(proc);
+                let bp_id = harness.add_breakpoint(bp_pc).unwrap();
+
+                harness.gdb_continue().unwrap();
+                std::thread::sleep(std::time::Duration::from_millis(500));
+                let stop_reason = harness.wait_for_stop_reason().unwrap();
```
yuv418, line 1004: same flaky question here

styx/plugins/styx-gdbserver/src/target_impl.rs
```diff
@@ -279,132 +285,90 @@ where
...
-                } else {
-                    // loop until there's a reason to exit
-                    loop {
-                        // poll for incoming data before every epoch
-                        if poll_incoming_data() {
-                            debug!("runtime poller found incoming data from gdb clien");
-                            break RunEvent::IncomingData;
-                        }
+        // continue until something pauses execution
+        // NOTE: if any watchpoints are set then we must single step
```
yuv418, line 289: Could mentioning this information be useful to add to styx docs on debugging, since presumably doing `continue` and having single-stepping occur could cause unexpected performance impact?

styx/plugins/styx-gdbserver/src/target_impl.rs
```diff
@@ -279,132 +285,90 @@ where
...
+        // because we need to check the memory state every single
+        // instruction
+        // Should we execute one insn at a time, or in large strides?
+        let should_step = self.exec_mode != ExecMode::Continue || !self.watchpoints.is_empty();
+        // Should we +1 to self.step_cycles?
+        // If we are in continue mode, we always should. This allows IRQs to happen if the user
+        // continues but we are stepping to catch watpoints.
+        // If we are in a step exec mode then this depends on the step irqs option.
+        let should_step_irqs =
+            self.exec_mode == ExecMode::Continue || self.options.step_irqs == StepIRQs::Enabled;
```
yuv418, line 299: Just to make sure I understand, does this line ensure that IRQs are triggered with `continue` even if `continue` is single stepping because there are memory watchpoints?

styx/plugins/styx-gdbserver/src/target_impl.rs
yuv418: I think vliw stands for "very long instruction word" 

Edit: it looks like the term is overloaded?

yuv418 (commented):
Overall looks good! 

The only "important" question I have is: if I want to change the number of instructions between each tick, if I want this to be consistent, then do I have to change it both with the `GDBOptions` and somewhere that would affect the "stride length" in the executor (is this changeable without doing another `ExecutorImpl`)? 

^^ That might be a different question for another day though. My thoughts are mainly about having fragmentation in where these values have to be defined.

styx/plugins/styx-gdbserver/src/options.rs
YurBoiRene: The GDBExecutor is special in that it doesn't really use the main `Executor` structure (see `styx/plugins/styx-gdbserver/src/plugin.rs:197` ish) so `post_stride_processing` is never called here.

yes you are correct though each executor impl would set the stride length. Does this answer your question? I know it is not ideal, I am open to suggestions!

styx/plugins/styx-gdbserver/src/target_impl.rs
YurBoiRene: Yeah I believe this was just a placeholder when we added wall clock times. I can fix it, seems easy.

styx/integration-tests/src/gdb_core_integration_test_suite.rs
YurBoiRene: Removing the sleep seems to make them flaky, looking into this for a better solution.

YurBoiRene: resolved by waiting for target to be running before waiting for stop.

styx/plugins/styx-gdbserver/src/target_impl.rs
YurBoiRene: added to `docs/source/workflows/debuggable.rst`, there was already a note there but I improved it.

styx/plugins/styx-gdbserver/src/target_impl.rs
YurBoiRene: Yes, if we are `continue`ing, then always do IRQ logic, if we are stepping, only do IRQ logic if enabled via config.

styx/plugins/styx-gdbserver/src/options.rs
yuv418: Yeah, that does answer the question. I'll take the discussion to the main PR comments.

styx/plugins/styx-gdbserver/src/target_impl.rs
YurBoiRene: I agree with you that it is confusing to use VLIW here. Removed the acronym

styx/plugins/styx-gdbserver/src/target_impl.rs
```diff
@@ -279,132 +285,94 @@ where
...
+        let mut last_tick = Instant::now();
+        loop {
+            if self.step_cycles >= self.options.cpu_epoch {
+                // insert interrupt
+                _ = self
+                    .proc
+                    .event_controller
+                    .next(self.proc.cpu.as_mut(), &mut self.proc.mmu);
+
+                let delta = Delta {
```
yuv418, line 309: I noticed the fuzzer plugin also seems to assume the 1 ns per instruction, maybe worth doing in another PR

yuv418: I also wonder if it's worth mentioning the assumptions for the Delta passed to tick in some place related to peripheral documentation

yuv418 (approved):

PR #168, opened 2026-01-08
doc: timers for users

### LINKED ISSUE #154: Add guidance on timers
Many processors need timer registers to be implemented. This is normally one of the first peripherals implemented and is integral for running an OS, but the correct way to implementing timers in Styx is not obvious. This should be a documentation page detailing the many different ways on implementing timers along with their tradeoffs.

## Methods to Implement Timers (as I remember)

- Peripheral tick
- In pc manager
- Read memory hook
- Read register hook
- Code hook

YurBoiRene:
closes #154

### DIFF
--- a/docs/source/user/timers.rst
+++ b/docs/source/user/timers.rst
@@ -1,0 +1,206 @@
+Timers
+======
+
+Timers are fundamental components commonly found in processors. A timer operates
+by periodically generating an interrupt request (IRQ) after a specified number
+of instructions have executed. These IRQs can typically be masked or configured
+to control their trigger frequency. In real-time operating systems (RTOS),
+timers are essential for task switching, while in general-purpose operating
+systems, they are used for time tracking and scheduling.
+
+Styx provides several approaches for implementing timers, each with different
+trade-offs between accuracy, performance, and implementation complexity.
+
+We distinuquish between target timer registers being **memory mapped I/O**
+for peripherals (e.g. arm) or **CPU registers** (e.g. PowerPC's ``TBL`` and
+``TBU``).
+
+A naive implementation of a timer IRQ may use an asyncronous timer that triggers
+every 100ms, or other constant duration. However, you will find that most of
+these implementations listed here rely on triggering timer IRQs after a constant
+number of instructions, rather than duration of time. This is to ensure target
+emulation is **deterministic** across executions. This is becase instruction
+cycles are not linked to a specific timestep so the host machines CPU speed and
+load will change how many instructions are executed in a given 100ms wall clock
+time interval from one run to the next. What this materializes into is that a
+bug may occur if a timer is triggered at a specific instruction type or in a
+critical point in a task. To investigate further, you rerun the test and the
+next run, your host machine is slightly faster and instead of triggering at the
+critical point, the timer is triggered at a benign instruction, thus missing
+the bug.
+
+Without deterministic execution, debugging becomes
+difficult. It is recommended your timer implementation follows this advice.
+
+See more about the :ref:`async-implementation` and the example processor that uses it below.
+
+Code Hook Implementation
+------------------------
+
+The first approach to implementing a timer in Styx is to use a Code Hook on
+every address. Code Hooks execute before the instruction at their registered
+address. By creating a code hook on every address using ``StyxHook::code(..,
+my_hook)``, the hook will trigger on every instruction. This hook can increment
+an internal counter and/or update a register (either memory-mapped or CPU
+register) to maintain the timer state. When the timer condition is met, the
+code hook has access to the ``EventController`` to latch the appropriate timer
+IRQ.
+
+**Pros:**
+
+1. Timer state is updated every instruction, ensuring hardware accuracy.
+2. Simple to implement.
+3. Works with all timer implementations (memory-mapped registers, CPU registers).
+
+**Cons:**
+
+1. Significant performance cost.
+
+**Examples:**
+
+1. ``ppc4xx`` / ``styx/processors/ppc/styx-ppc4xx-processor/src/timers.rs``
+
+Tick-Based Implementation
+-------------------------
+
+The second approach is to update timers in the ``tick()`` function of a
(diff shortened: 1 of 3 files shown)


docs/source/user/timers.rst
```diff
@@ -0,0 +1,176 @@
...
+
+We distinuquish between target timer registers being **memory mapped I/O**
+for peripherals (e.g. arm) or **CPU registers** (e.g. PowerPC's ``TBL`` and
+``TBU``).
+
+A naive implementation of a timer IRQ may use an asyncronous timer that triggers
+every 100ms, or other constant duration. However, you will find that most of
+these implementations listed here rely on triggering timer IRQs after a constant
+number of instructions, rather than duration of time. This is to ensure target
+emulation is **deterministic** accross executions. Otherwise, debugging becomes
```
yuv418: Could it be helpful to link to the sample timer that does this asynchronous timing? I think you have it below, saying (see below for example) or linking twice could be useful IMO

Also, it might be useful to explain why the asynchronous timer is not deterministic (in my head I was thinking if emulation is slower than the actual processor, then using wall clock time for our timer won't reflect the the actual "progress" made in terms of instructions executed compared to having the timer tick every fixed number of instructions).

docs/source/user/timers.rst
```diff
@@ -0,0 +1,176 @@
...
+**Examples:**
+
+1. ``ppc4xx`` / ``styx/processors/ppc/styx-ppc4xx-processor/src/timers.rs``
+
+Tick-Based Implementation
+-------------------------
+
+The second approach is to update timers in the ``tick()`` function of a
+peripheral or event controller. This method is similar to the code hook approach
+but executes less frequently, usually once every 1000 instructions. At every
```
yuv418: > usually once every 1000 instructions

A question that comes up when I read this is "when is it not 1000 instructions" or how can I change this?

docs/source/user/timers.rst
```diff
@@ -0,0 +1,176 @@
...
+``tick()``, the Event Controller or Peripheral will write the CPU or Memory
```
yuv418: Related to previous comment, but maybe mentioning that the `tick` function is given the number of instructions from the last tick to now (at least if I understand it correctly, this is what happens) might be useful

docs/source/user/timers.rst
```diff
@@ -0,0 +1,176 @@
...
+
+1. Hardware accuracy is maintained to a reasonable degree (registers are updated).
+2. Works with all timer implementations (memory-mapped registers, CPU registers).
+3. More performant than the code hook approach.
+
+**Cons:**
+
+1. Performance cost from register/memory operations approximately every 1000 instructions.
+2. Target timer state not up to date between ticks.
+3. Requires scaling logic for tick-based implementation.
```
yuv418, line 97: 🙂 (the scaling logic is definitely something I had to deal with)

docs/source/user/timers.rst
```diff
@@ -0,0 +1,176 @@
...
+Tick-Based Implementation
+-------------------------
+
+The second approach is to update timers in the ``tick()`` function of a
+peripheral or event controller. This method is similar to the code hook approach
+but executes less frequently, usually once every 1000 instructions. At every
+``tick()``, the Event Controller or Peripheral will write the CPU or Memory
+Mapped timer register with the current timer state. Then the same system of
+code can trigger the timer IRQ as configured. This may involve reading timer
+configuration registers or Memory Write hooking timer configuration registers
```
yuv418: nitpick: I feel like "Memory Write hooking.." could maybe be rephrased/elaborated since it feels a bit confusing to parse for me. Maybe something like "adding a memory write hook over the MMIO region" (not 100% if that phrasing is best).

docs/source/user/timers.rst
```diff
@@ -0,0 +1,176 @@
...
+**Cons:**
+
+1. Only applicable to CPU timer registers.
+2. System status may not be up to date between ticks.
+3. Requires scaling logic for tick-based implementation.
+4. Timers are only updated via memory hooks, so GDB and Styx ``read_memory()`` operations will not reflect current timer register status.
+5. Requires the Pcode Backend.
+
+PC Manager Implementation
+--------------------------
```
yuv418, line 157: Are there examples of this?

yuv418 (changes requested):
Very very helpful doc that will be good to refer to for future timer implementers (:

In terms of suggestions, my comments are mostly just: some explanations on the deterministic stuff could be helpful, and maybe some random rewording. 

An overall thought was also I wonder if it's worth saying that you _can_, in some sense, "mix and match" some of these approaches (eg. if your arch has both register and MMIO, you could theoretically hook the registers but write memory in the tick). I sort of did this for Hexagon.

Another thing maybe (I am guessing this would go in a different doc about peripherals) would be to discuss more generally what "hooking writes for MMIO registers" looks like in terms of code.

docs/source/user/timers.rst
YurBoiRene: changed to "In order to configure the timer, this method may involve reading timer
configuration registers periodically or creating a Memory Write hook on the
MMIO registers used to configure the timers."

docs/source/user/timers.rst
YurBoiRene: No

docs/source/user/timers.rst
```diff
@@ -0,0 +1,205 @@
...
+**Examples:**
+
+1. ``kinetis21`` / ``styx/processors/arm/styx-kinetis21-processor/src/systick.rs``
+    1. Note: this implementation does not write to the actual register, it only triggers IRQs.
+
+.. [#note1]
+   The number of instructions between ``tick()`` is called the "stride
+   length". It is constant throughout emulation and statically defined in
+   the ``ExecutorImpl::get_stride_length()``. This value is 1000 for the
+   ``DefaultExecutor``. The exeception is the ``GdbExecutor`` which uses custom
```
yuv418, line 109: Is it worth making it more explicit that it is possible to change the stride length?

PR #169, opened 2026-01-16
Concurrent Mmu

### LINKED ISSUE #159: Move MMU/TLB to Live Next to Cpu
Each vCPU should own a `CpuBackend` and `TlbImpl`, the `TlbImpl` will probably be contained in an `Mmu` along with a shared reference to `MainMemory`.. The `TlbImpl` and `Mmu` will interface with `&MainMemory`. 

```
┌─────────────────────────────────────┐
│ vCPUs x1, x2, xN                    │
│┌───────────────────────────────────┐│
││ CpuBackend           Mmu          ││
││┌───────────┐        ┌────────────┐││
│││Registers  │        │TlbImpl     │││
│││           │        │            │││
│││Cpu State  │        │&MainMemory │││
│││           │        │   │        │││
││└───────────┘        └───┼────────┘││
│└─────────────────────────┼─────────┘│
│                          │          │
│                          │          │
│                          │          │
│                          ▼          │
│                       MainMemory    │
│                                     │
└─────────────────────────────────────┘
```

YurBoiRene:
Big Changes:
- `Mmu` now contains `Arc<MemoryBackend>` to allow for concurrent access to physical memory, while each vCPU gets a Mmu. 
- This change removes the ability to add memory regions during emulation
  - We might be able to remedy this in the future by adding something like an [append-only-vec](https://github.com/droundy/append-only-vec/) but I wanted to avoid putting a mutex on the memory regions.

A lot of the work was modifying `ProcessorImpl`s to use the new API. I am thinking of providing some helper methods that would remain consistent to avoid widespread changes if we change how memory is initialized again.

Closes #159

### DIFF
--- a/styx/bindings/styx-py-api/pyproject.toml
+++ b/styx/bindings/styx-py-api/pyproject.toml
@@ -20,4 +20,4 @@ profile = "release"
 auditwheel = "check"
 strip = true
 sdist-generator = "cargo"
-python-source = "python/styx_emulator"
+python-source = "python"


styx/core/styx-cpu-pcode-backend/tests/basic_blocks.rs
```diff
@@ -28,13 +29,13 @@ fn test_basic_block_hook() -> Result<(), UnknownError> {
     14:	4b ff ff ec 	b       0 <start>
    ";
 
-    let mut mmu = Mmu::default_region_store();
+    let mut mmu =
+        Mmu::with_regions([MemoryRegion::new(0, 0x10000, MemoryPermissions::all()).unwrap()]);
```
yuv418, line 33: self-note: is it a "lot of work" for someone to initialize an `Mmu` now, or does `default_region_store` still exist and there is a better reason to use `with_regions`? Investigate this

yuv418: self-note to also understand the difference between `MemoryBackend` and `Mmu` more.

styx/core/styx-cpu-pcode-backend/tests/hexagon_tcg.rs
```diff
@@ -92,17 +97,23 @@ fn test_binutils_unittests(test: TestData) {
         HexagonVariants::QDSP6V66,
         ArchEndian::LittleEndian,
     );
-    let mut mmu = Mmu::default_region_store();
+    let mut memory = MemoryBackend::new_region_store();
+    let tlb = dummy_tlb();
```
yuv418: self-note: tlb seems to be part of mmu now

styx/core/styx-cpu-pcode-backend/tests/ummapped_execution.rs
```diff
@@ -75,17 +72,15 @@ fn test_goes_unmapped_execution() {
...
+    let mut mmu = Mmu::with_regions([
+        MemoryRegion::new(0, 0x1000, MemoryPermissions::all()).unwrap(),
+        MemoryRegion::new(0x1000, 0x1000, MemoryPermissions::empty()).unwrap(),
+    ]);
     let mut ev = EventController::default();
 
     let mut cpu =
         PcodeBackend::new_engine(Arch::Ppc32, Ppc32Variants::Ppc405, ArchEndian::BigEndian);
 
-    mmu.add_memory_region(MemoryRegion::new(0x0, 0x1000, MemoryPermissions::all()).unwrap())
```
yuv418, line 84: self-note: perhaps the same amount of work for memory region.

styx/core/styx-cpu/tests/comparison_tests.rs
```diff
@@ -524,25 +506,14 @@ fn test_compare_simple_read_memory_hook() {
...
-    let mut machines = create_machines(
+    let mut machines = create_machines_regions(
         "movs r0, #0x00
          movs r1, #0xDE
          str r1, [r0]
          movs r1, #0xAA
          str r1, [r0, #20]",
-    );
-
-    // map memory at 0x00
```
yuv418, line 535: Why did this get removed?

yuv418: Does it map memory during emulation?

styx/core/styx-processor/src/core/mod.rs
```diff
@@ -81,17 +80,8 @@ impl ProcessorCore {
     pub fn dummy() -> Self {
         Self {
             cpu: Box::new(DummyBackend),
-            mmu: Mmu::from_impl(Box::new(DummyTlb)),
```
yuv418, line 84: What did the `DummyTlb` do here when passed to the `Mmu`?

styx/core/styx-processor/src/core/mod.rs
```diff
@@ -81,17 +80,8 @@ impl ProcessorCore {
...
+            mmu: Mmu::default(),
             event_controller: EventController::new(Box::new(DummyEventController::default())),
         }
     }
-
-    /// Repackage the [`ProcessorCore`] as a [`CoreHandle`] struct for use within hooks.
-    pub fn core_handle(&mut self) -> CoreHandle {
```
yuv418, line 90: self-note: investigate how the CoreHandle was changed in general

styx/core/styx-processor/src/event_controller/mod.rs
```diff
@@ -181,10 +189,14 @@ impl EventController {
         }
     }
 
-    pub fn reset(&mut self, cpu: &mut dyn CpuBackend, mmu: &mut Mmu) -> Result<(), UnknownError> {
-        self.inner.reset(cpu, mmu)?;
+    pub fn reset(
+        &mut self,
+        cpu: &mut dyn CpuBackend,
+        memory: &mut Mmu,
```
yuv418, line 195: should this also be MemoryBackend?

styx/core/styx-processor/src/hooks/core_handle.rs
```diff
@@ -90,7 +90,6 @@ impl<'a> CoreHandle<'a> {
             pub fn code(&mut self) -> crate::memory::CodeMemoryOp;
             /// See [`Mmu::data()`]
             pub fn data(&mut self) -> crate::memory::DataMemoryOp;
-            /// See [`Mmu::sudo_code()`]
```
yuv418, line 93: should this comment come back? all the other comments still exist

styx/core/styx-processor/src/memory/tlb/dummy.rs
```diff
@@ -8,9 +8,21 @@ use crate::memory::{
     MemoryOperation, TlbImpl, TlbTranslateError,
 };
 
+/// Create a TLB that doesn't perform any translation on addresses.
+///
+/// Box'd for convenience.
+pub fn dummy_tlb() -> Box<DummyTlb> {
```
yuv418: What's the difference between this and `DummyTlb::new`? It looks like they do the same thing, so would it be easier to just have one and use it everywhere? I might be missing something, though

yuv418: Also if I had to choose one, I would probably choose `DummyTlb::new`, but not sure.

styx/core/styx-processor/src/memory/memory_region.rs
```diff
@@ -236,7 +237,10 @@ impl RegionData {
 ///
 /// Comparison's between [`MemoryRegion`]'s do not compare data,
 /// only the addresses and sizes. When comparing regions it is
-/// assumed that the regions do *NOT** overlap
+/// assumed that the regions do *NOT** overlap.
+///
+/// [`MemoryRegion`] cloned with `.clone()` will copy the underlying data.
```
yuv418, line 242: this answers my question

styx/core/styx-processor/src/memory/physical.rs
```diff
@@ -143,9 +143,38 @@ pub enum PhysicalMemoryVariant {
...
+/// For this reason, avoid zeroing large sections of memory when using the Flat Memory.
+///
+/// Harvard configuration splits code and data memory while VonNeumann has a unified
+/// memory space.
+/// Read/write operations have `_code` and `_data` variants which will be identical
+/// in a VonNeumann configuration.
+/// [`MemoryBackend`]
+///
+/// The [`MemoryBackend`] is designed for concurrent memory reads and writes.
+/// Memory regions require `&mut` and so should be added during construction
```
yuv418: I feel like this comment could be refactored/expanded on to be more clear. Maybe more info on why you can only remove/add regions during construction

styx/core/styx-processor/src/memory/mmu.rs
```diff
@@ -43,59 +42,73 @@ pub enum MemoryType {
...
 }
 
-/// The MMU owns the processor specific TLB implementation and the physical memory backend. It is
-/// the single point where all memory transactions flow through.
+/// Per-cpu MMU allows access to virtual and physical memory.
+///
+/// The MMU has two components: the processor specific TLB implementation and
+/// a reference to the physical memory backend.
+/// It is the single point where all memory transactions should flow through.
+/// The Mmu has full mutable access to the [`TlbImpl`] but an **immutable** reference to the [`MemoryBackend`] allowing it to
```
yuv418: self-note: investigatre how this allows memory modification

styx/core/styx-processor/src/memory/mmu.rs
```diff
@@ -43,59 +42,73 @@ pub enum MemoryType {
...
-    ) -> Result<Self, UnknownError> {
-        tlb.init(cpu)?;
-        Ok(Self {
-            tlb,
-            memory: MemoryBackend::new(memory),
-        })
-    }
+    /// Construct a default Mmu with a set of regions.
+    ///
+    /// This is useful for testing or processor that do not have a custom TLB.
```
yuv418, line 88: I feel like this comment could say something more explicitly about the fact that the dummy tlb is used. A question is: if I have my own tlb that I want to use, could it be useful to have a similar method that lets me provide a tlb of my choosing?

styx/core/styx-processor/src/memory/mmu.rs
```diff
@@ -43,59 +42,73 @@ pub enum MemoryType {
...
-    /// Constructs the [`Mmu`] with a [`DummyTlb`] and a [`PhysicalMemoryVariant::RegionStore`].
-    pub fn default_region_store() -> Self {
-        // OKAY to subvert ::new() because DummyTlb has noop init.
-        Mmu {
-            tlb: Box::new(DummyTlb),
-            memory: MemoryBackend::new(PhysicalMemoryVariant::RegionStore),
+    /// Create a new Mmu with dummy tlb and flat memory.
+    ///
+    /// This is good for testing or proof of concept processors.
+    /// This is not recommended for production processors since there are no memory regions.
```
yuv418: is there more documentation on why having no memory regions is bad?  i feel like saying "has flat memory" here could also be useful

styx/core/styx-processor/src/memory/mmu.rs
yuv418: (slash add more comment/doc to make this clearer)

styx/core/styx-cpu/tests/comparison_tests.rs
YurBoiRene: `create_machines_regions` maps the regions, yes. Let me add some docs

styx/core/styx-processor/src/core/mod.rs
YurBoiRene: Nothing, The default also uses teh DummyTlb

styx/core/styx-processor/src/hooks/core_handle.rs
YurBoiRene: Good catch, readded

styx/core/styx-processor/src/memory/mmu.rs
YurBoiRene: Documented and added other method for this use case

styx/core/styx-processor/src/memory/mmu.rs
YurBoiRene: Added more documentation

styx/core/styx-processor/src/memory/mmu.rs
YurBoiRene: Any changes needed here?

yuv418: Oh, I missed that:

> but an **immutable** reference to the [`MemoryBackend`] allowing it to modify memory concurrently with other vCPUs.

I think this line might have confused me, because it wasn't clear how to go from "immutable reference" to "modify memory concurrently" (I think I looked into it and maybe there was some interior mutability but whatever it was, I think it would be helpful to say here)

YurBoiRene: Reworded

styx/core/styx-processor/src/memory/physical.rs
```diff
@@ -143,9 +143,46 @@ pub enum PhysicalMemoryVariant {
...
+/// Read/write operations have `_code` and `_data` variants which will be identical
+/// in a VonNeumann configuration.
+/// [`MemoryBackend`]
+///
+/// ## Concurrency
+/// The [`MemoryBackend`] is designed for concurrent memory reads and writes.
+/// Concurrent reads and writes to memory are enabled by using host Atomic reads and writes.
+/// This way memory operations do not require a lock and can be fast.
+///
+/// While the memory itself can be modified using atomics, adding and removing regions
```
yuv418, line 183: https://stackoverflow.com/questions/35810843/why-do-rusts-atomic-types-use-non-mutable-functions-to-mutate-the-value 

I probably should have thought of it (obviously after reading this explanation everything makes sense), but I guess if the question exists then maybe it's worth saying atomic operations are immutable?

styx/core/styx-processor/src/memory/mmu.rs
```diff
@@ -43,59 +41,99 @@ pub enum MemoryType {
...
+    /// Construct a Mmu with a set of regions and default TLB.
+    ///
+    /// This is useful for testing or processor that do not have a custom TLB.
+    pub fn with_regions(regions: impl IntoIterator<Item = MemoryRegion>) -> Self {
+        Self::with_regions_tlb(regions, DummyTlb::new())
     }
 
-    /// Constructs the [`Mmu`] with the default physical memory backend.
-    pub fn from_impl(tlb: Box<dyn TlbImpl>) -> Self {
+    /// Construct a Mmu with a set of regions and custom TLB.
```
yuv418, line 93: Maybe mention "dummy tlb" here instead of "custom tlb" as well?

yuv418 (changes requested):
just two spots where I think docs could maybe still be polished a bit (to some degree maybe nitpicking, so feel free to ignore and I can approve), other than that it's good

styx/core/styx-processor/src/memory/physical.rs
YurBoiRene: How is this?


```
/// The [`MemoryBackend`] is designed for concurrent memory reads and
/// writes. Concurrent reads and writes to memory are enabled by using host
/// [`std::sync::Atomic`] reads and writes. Because Atomic operations are
/// immutable (only requiring `&`), we can modify memory from multiple vCPUs,
/// threads, etc all without requiring a lock on memory, which would cause a
/// major performance hit.
```

yuv418: LGTM! :+1:

PR #174, opened 2026-02-06
fix: write register fails on incorrect size

YurBoiRene:


styx/core/styx-cpu-pcode-backend/src/lib.rs
```diff
@@ -470,6 +470,12 @@ impl CpuBackend for PcodeBackend {
         if reg == pc_reg_variant {
             self.set_pc(sized_value.to_u64().with_context(|| "too big")?)?;
         } else {
+            if !reg.register_value_enum().size_equal(value) {
```
yuv418: Should this also work for the pc? On hexagon for example, the pc is 32 bits.

yuv418: Oh, I guess it doesn't have to since the PC is internally stored as 64 bits. (?)

styx/core/styx-cpu-pcode-backend/src/register_manager.rs
```diff
@@ -185,9 +185,20 @@ impl<T: CpuBackend> RegisterManager<T> {
         cpu: &mut dyn RegisterCallbackCpu<T>,
         register: ArchRegister,
         value: SizedValue,
-    ) -> Result<(), RegisterHandleError> {
+    ) -> Result<(), WriteRegisterError> {
         trace!("Triggering write_register index {register}.",);
 
+        let pc_reg_variant = cpu.pc_register().variant();
+        if register != pc_reg_variant
```
YurBoiRene: Wouldn't this mean that a u64 write to pc will "overflow" in the pcode memory, just as you were seeing before? Is there a reason we need this exception.

I did notice in the normal pcode backend, set_pc() is wrong in that it truncates to 4 bytes, when it should truncate to the size of the pc register. Maybe we need to change that?

yuv418: The reason I added the exception is because in the original implementation in PcodeBackend I noticed that you only did the check if it wasn't the PC register, so I tried to keep it consistent. I'm not sure what the correct behavior is here, but I feel like a silent truncation could also lead to footguns.

yuv418: Also, I think one of the reasons why might also be because I see a lot of places where the PC seems to be hardcoded to 64-bits, and there were issues with that in some of the test cases.

YurBoiRene: Is that in a `write_register(PC, n)` or in a `set_pc()`?

I guess to make the functionality be the same there should be an exception but it should probably do something like this.

```

let pc_reg_variant = self.pc_register().variant();
let pc_size = pc_reg_variant.size();
let isa_pc = SizedValue::from_u128(input_pc as u128,  pc_size);
RegisterManager::write_register(self, pc_reg_variant, isa_pc)?;

```

yuv418: It was in `set_pc`. For example, at `styx/processors/aarch64/styx-aarch64-processor/src/lib.rs:104`.

yuv418: I'll try that

YurBoiRene: Looks like it worked? heck yeah

yuv418: Yeah I think it did!

PR #192, opened 2026-04-27
Add l2vic and qtimer support

appar3ntly:
Adding on behalf of @yuv418 's work as we sort out upstreaming hexagon stuff.

* Adds l2vic and qtimer support sufficient to pass relevant qemu-hexagon-testing tests.

There will probably be some overlap/conflicts with the other upstreaming PRs. We can at least resolve them once and likely have identical changes across them.

### Note on ordering

Because I don't want to redo all of the patch creation, this has a commit that removes exception.rs and its module declaration for hexagon. It looks like it's actually unused as of the final hexagon upstreaming changes.

We just need to make sure that source/declaration are kept **in** throughout conflict resolutions.

### Testing

(moving to separate PR to add testing infra)

~~You can use the new `styx-hexagon-testing` crate in `incubation/` to test this.~~

~~1. You'll need the (compiled) tests from https://github.com/qualcomm/qemu-hexagon-testing , I can help with this if you don't already have them.~~
~~2. Put them say in your `styx-hexagon-testing` crate source.~~
~~3. `TEST_CASE=1 RUST_LOG=none cargo run --release -- qtests/qtimer_test`~~

~~I think `qtimer` and `qtimer_test` are the relevant timer tests.~~

Some notes on testing:

* As of writing you have to run as release, there's some under/overflow stuff that gets caught in debug that I assume needs to be allowed for these register ops.
* I haven't run every test, and I don't think they all will/should pass at this point, @yuv418 will know more.

PR #196, opened 2026-05-05
feat: QEMU hexagon tests, hexagon testing incubator 

yuv418:
# Testing 

(from #192)

- New styx-hexagon-testing crate for specifying targets, hooks, and config. 
- Processor config. I modified the EventController to expose this. The goal is to eventually eliminate exposing the config table to `HexagonDevice` (`cfgtable.rs`), and move everything to processor config. The HexagonBuilder will eventually set all the config table entires based on processor config, but this hasn't happened yet. The information set through processor config is not in the config table enum.
- New CI task for running hexagon tests. Includes an addition to the `Justfile` two shell scripts: one in the `styx-hexagon-testing` to grep `qemu-hexagon-tests`'s stdout to check if things have passed or failed, and one in the `styx-hexagon-testdata` to download/build `qemu-hexagon-tests` on the CI runner. It seems like a lot of overhead to use a container on the CI node, so the shell script exists to avoid that.
- There are some extra files in `styx-hexagon-testdata` that are WIPs on building the tests without the Hexagon SDK and one that creates a container image with the tests.

Some notes on testing:

- As of writing you have to run as release, there's some under/overflow stuff that gets caught in debug that I assume needs to be allowed for these register ops. (it works in `--release`, so that needs to be fixed)
- (from Kenneth) I haven't run every test, and I don't think they all will/should pass at this point, 
- There aren't any tests at the moment. I'll add those in the feature-related PRs; this is just the infrastructure for that testing.

Also, I don't know if shell scripts are the best way to do this, but we need a way to capture stdout from the test harness in order to tell if things are passing.


data/test-binaries/styx-hexagon-testdata/data/hexagon-builder.Containerfile
```diff
@@ -29,53 +29,81 @@ RUN apt-get update && \
...
 RUN DEBIAN_FRONTEND=noninteractive eatmydata \
     apt-get install -y --no-install-recommends \
         make \
+	build-essential \
         git \
         # why need this
-        libxml2
+        libxml2 \
+	cmake \
+	ninja-build
```
YurBoiRene: nit: inconsistent tabbing here?

data/test-binaries/styx-hexagon-testdata/data/hexagon-builder.Containerfile
```diff
@@ -29,53 +29,81 @@ RUN apt-get update && \
...
-    cp ../qemu/tests/tcg/hexagon/* src/
-COPY data/Makefile .
-RUN mkdir -p build/ && make -j $jobs all
+#RUN mkdir -p tests/
+#WORKDIR tests/
+#RUN mkdir -p src/ && \
+#    cp ../qemu/tests/tcg/hexagon/* src/
+#COPY data/Makefile .
+#RUN mkdir -p build/ && make -j $jobs all
+
```
YurBoiRene: Can probably delete all this?

data/test-binaries/styx-hexagon-testdata/data/hexagon-builder.Containerfile
```diff
@@ -29,53 +29,81 @@ RUN apt-get update && \
...
+  cmake --build build-systests
 
+# test artifacts are in /testdata/bin
 FROM build AS release
-RUN mkdir -p /testdata/bin
+RUN mkdir -p /testdata/bin/qemu-tests
+RUN mkdir -p /testdata/bin/qemu-hexagon-testing
 WORKDIR /testdata
-COPY --from=build /src/tests/build/* bin/
+#COPY --from=build /src/tests/build/* bin/qemu-tests
```
YurBoiRene: delete as well

data/test-binaries/styx-hexagon-testdata/build_ci.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
YurBoiRene: Nit: `#!/usr/bin/env bash` for n*xos users 🤢

data/test-binaries/styx-hexagon-testdata/build_ci.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# for CI
+
+# https://stackoverflow.com/questions/59895/how-do-i-get-the-directory-where-a-bash-script-is-located-from-within-the-script
```
YurBoiRene: Probably don't need the link to the solution, just put a comment on what the line does, e.g. get full path of script.

data/test-binaries/styx-hexagon-testdata/data/build_ci.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+# for CI
```
YurBoiRene: Explain *why* this is for CI. Looks like it caches the qemu-hexagon-testing clone on the runner and does not use the container.

data/test-binaries/styx-hexagon-testdata/build_ci.sh
```diff
@@ -0,0 +1,21 @@
...
+	git clone https://github.com/qualcomm/qemu-hexagon-testing $SCRIPT_DIR/qemu-hexagon-testing
+
+	cd $SCRIPT_DIR/qemu-hexagon-testing
+
+	cmake -S standalone_systests -B build-systests \
+	  -G Ninja \
+	  -DCMAKE_TOOLCHAIN_FILE=${PWD}/cmake/hexagon-standalone.cmake \
+	  -DHEXAGON_SDK_ROOT=$HEXAGON_SDK_ROOT \
+	  -DHEXAGON_ARCH=v71
+	cmake --build build-systests
```
YurBoiRene: Probably version pin this so that new tests added to qemu-hexagon-testing don't cause confusing CI failures.

incubation/styx-hexagon-testing/src/devices/mod.rs
YurBoiRene, line 1: Any reason not to use `UnknownError` instead of `Box<dyn std::error::Error>`?

styx/core/styx-macros/src/processor_config.rs
```diff
@@ -10,7 +10,7 @@ pub fn derive_processor_config(input: TokenStream) -> TokenStream {
 
     let mut trait_path = crate::styx_manifest::StyxManifest::shared(|m| m.get_processor_path());
     trait_path.segments.push(format_ident!("processor").into());
-    trait_path.segments.push(format_ident!("config").into());
+    // trait_path.segments.push(format_ident!("config").into());
```
YurBoiRene: Is this needed for this to work?

data/test-binaries/styx-hexagon-testdata/data/build_ci.sh
YurBoiRene, line 1: Run shellcheck on this and fix the comments. The biggest one I see is that if the script is in a path with a space somewhere in the directory tree, it will split in the places where `$SCRIPT_DIR` is not quoted.

justfile
```diff
@@ -176,3 +176,7 @@ docker *ARGS:
 # Pass all arguments to cargo
 cargo *ARGS:
     cargo {{ARGS}}
+
+hexagon-test:
```
YurBoiRene: What is the recommended way to run on local, not cI? Add some docs or a script to build the hexagon tests local.

data/test-binaries/styx-hexagon-testdata/data/hexagon-builder.Containerfile
YurBoiRene, line 1: Update the docs at the top of this file. It is no longer fetching the linaro toolchain

incubation/styx-hexagon-testing/tester.sh
```diff
@@ -0,0 +1,21 @@
...
+
+if [ ! -d "$HEXAGON_TESTS_LOCATION" ]; then
+	echo "$HEXAGON_TESTS_LOCATION does not exist. Please run ./build_ci.sh."
+	exit 1
+fi
+
+# Running as debug will expose more errors
+for test in "${TESTS[@]}"
+do
+	RUST_LOG=info cargo run --manifest-path=$SCRIPT_DIR/Cargo.toml -- --target tester $HEXAGON_TESTS_LOCATION/${test} | grep PASS
```
YurBoiRene: Same `SCRIPT_DIR` could have spaces comment, double quote

incubation/styx-hexagon-testing/tester.sh
```diff
@@ -0,0 +1,21 @@
...
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+declare -a TESTS=()
+
+if [ ! -d "$HEXAGON_TESTS_LOCATION" ]; then
+	echo "$HEXAGON_TESTS_LOCATION does not exist. Please run ./build_ci.sh."
+	exit 1
+fi
+
+# Running as debug will expose more errors
```
YurBoiRene: What does this mean?

incubation/styx-hexagon-testing/tester.sh
YurBoiRene: Does the tester target ever print "PASS"? or it just a stub for later?

data/test-binaries/styx-hexagon-testdata/data/build_ci.sh
YurBoiRene, line 1: Also update the readme in this directory.

YurBoiRene (changes requested):

data/test-binaries/styx-hexagon-testdata/build_ci.sh
yuv418: True

data/test-binaries/styx-hexagon-testdata/data/build_ci.sh
yuv418: I don't use the container in the CI; the container scripts are actually unused. I wanted to leave them so that people might be able to build a local container and use it instead of setting up the SDK and tests from scratch.

styx/core/styx-macros/src/processor_config.rs
yuv418: Yes, but I should delete the commented out line. Without doing this, I get import visibility issues

justfile
yuv418: I'll integrate the container for that

data/test-binaries/styx-hexagon-testdata/data/build_ci.sh
yuv418: this is a pretty cool tool. TIL!

incubation/styx-hexagon-testing/tester.sh
yuv418: It's not the tester that prints pass, but rather the test case programs.

yuv418:
For styx-hexagon-testdata:

- Use `build.rs` in `styx-hexagon-testdata` to build stuff using docker like it was before
- Take the old `hexagon_tcg.rs` test from `styx-cpu-pcode-backend`, adapt it into something we can use in `styx-hexagon-testing`. This will require
  - [x] Elf loading from binary files, not just file names. `main()` logic split into a common function for tests and main runner.
  - [x] Move semihosting outputs to be through a processor-config shared semihosting output mpmc channel.

incubation/styx-hexagon-testing/tester.sh
yuv418: This was kind of silly, but RUST_LOG=debug might have more error messages. I really just wanted to see if the TargetExitReason was bad since I had errors with that. Luckily that is fixed now and we can check that in a wayyyy less non hacky way (:

incubation/styx-hexagon-testing/src/devices/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    fn proc_config(&self) -> Result<HexagonProcessorConfig, UnknownError> {
+        self.proc_config()
+    }
+
+    fn post_init(&self, proc: &mut Processor) -> Result<(), UnknownError> {
+        self.post_init(proc)
+    }
+}*/
```
YurBoiRene: Delete commented code

incubation/styx-hexagon-testing/src/devices/pixel5.rs
```diff
@@ -0,0 +1,71 @@
...
+        Ok(vec![
+            StyxHook::MemoryRead(
+                (0x04080028..0x0408002c).into(),
+                Box::new(
+                    |proc: CoreHandle, _address: u64, _size: u32, _data: &mut [u8]| {
+                        proc.mmu.write_u32_le_phys_data(0x04080028, 0xfffffff0)?;
+
+                        Ok(())
+                    },
+                ),
```
YurBoiRene, line 29: Can you document the magic numbers in this file? I am not sure what these addresses and values are for.

incubation/styx-hexagon-testing/src/devices/tester.rs
```diff
@@ -0,0 +1,55 @@
...
+    errors::UnknownError,
+    hooks::StyxHook,
+    prelude::Processor,
+    processors::hexagon::hexagon::{HexagonConfigTable, HexagonProcessorConfig},
+};
+
+use super::HexagonDevice;
+
+#[derive(Default)]
+pub struct Tester {}
```
YurBoiRene, line 19: Maybe document when you would use this HexagonDevice?

incubation/styx-hexagon-testing/src/devices/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+    Tester,
+}
+
+pub trait HexagonDevice {
+    // Hooks
+    fn hooks(&self) -> Result<Vec<StyxHook>, UnknownError>;
+    // Set the processor config
+    fn proc_config(&self) -> Result<HexagonProcessorConfig, UnknownError>;
+    // To do fixups. Is sort of unsafe, since it could overwrite anything of the above, which isn't cool.
+    // Runs after the previous three functions.
```
YurBoiRene: "Previous three functions"? I assume this is just leftover when there was 3 above.

incubation/styx-hexagon-testing/src/main.rs
```diff
@@ -0,0 +1,188 @@
...
+        ),
+    ))?;*/
+
+    /*proc.core.cpu.add_hook(StyxHook::RegisterWrite(
+        HexagonRegister::BadVa0.into(),
+        Box::new(|proc: CoreHandle, reg, data: &RegisterValue| {
+            warn!("badva0 written pc {:x?}", proc.cpu.pc());
+            Ok(())
+        }),
+    ))?;*/
```
YurBoiRene: Maybe add some cli flags to enable these or remove? Not sure how useful/verbose they are.

incubation/styx-hexagon-testing/src/main.rs
```diff
@@ -0,0 +1,188 @@
...
+                warn!(
+                    "memory write hook at {:x?}, address is {address:x} lr {lr:x} value is {data:?} sz {size}",
+                    proc.pc()
+                );
+                Ok(())
+            },
+        ),
+    ))?;
+
+    // proc.core.mmu.write_u32_le_phys_data(0xf80044, 0x100)?;
```
YurBoiRene: Document or remove

incubation/styx-hexagon-testing/src/runner.rs
```diff
@@ -0,0 +1,70 @@
...
+    if let Some(debug) = debug {
+        let gdb_params = GdbPluginParams::tcp("0.0.0.0", debug.gdb_remote_port, true);
+        proc = proc.with_executor(
+            GdbExecutor::<HexagonHvxCpuTargetDescription>::new(gdb_params)?.with_options(
+                GDBOptions {
+                    step_irqs: StepIRQs::Enabled,
+                    ..Default::default()
+                },
+            ),
+        )
```
YurBoiRene, line 53: nit: you could print out the debugger connection params just to be nice.

"GDB listening on 0.0.0.0:port"

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/tests/arith.rs
YurBoiRene, line 1054: nit: i've been liking putting `println` in tests so a naive `cargo nextest run --package <package> --nocapture` shows the test specific output.

Curious what you think.

styx/processors/hexagon/styx-hexagon-processor/src/cfgtable.rs
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: BSD-2-Clause
+
```
YurBoiRene, line 2: Put a comment here: "The u32 repr is the offset into the cfgtable"

.github/workflows/hexagon.yml
```diff
@@ -0,0 +1,73 @@
+name: hexagon-ci
+on:
+  pull_request:
+    # dont run if only docs change
+    paths:
+      - "data/test-binaries/styx-hexagon-testdata/**"
+      - "incubation/styx-hexagon-testing/**"
+      - "styx/core/styx-cpu-type/hexagon/**"
+      - "styx/core/styx-cpu-pcode-backend/**/hexagon/**"
+      - "styx/processors/hexagon/styx-hexagon-processor/**"
```
YurBoiRene, line 9: I feel like this should run on more paths than just these. For example, a change to styx-processor could break these tests.

YurBoiRene (changes requested):

incubation/styx-hexagon-testing/src/runner.rs
yuv418: I think it already prints this, but not through me (eg. the gdbserver crate does it)

yuv418: Yeah, it does. See `styx/plugins/styx-gdbserver/src/event_loop.rs`, lines 219/268. It uses `eprintln!` so it doesn't depend on logging either.

.github/workflows/hexagon.yml
yuv418: I would have liked to do a proper query on a dependency graph, but since that takes a while I just included the entirety of styx-core and kept the rest the same. I feel like that would be acceptable (as in, I'm overapproximating instead of underapproximating), but not sure. I did try to do some dependency queries and I did get a list of dependencies but all of them were in styx-core except hexagon-testdata, hexagon-testing, and hexagon-processor, which are already in here.

incubation/styx-hexagon-testing/src/main.rs
yuv418: I think maybe I want to interface with python here as this is kind of a playground thing and remove these

styx/core/styx-cpu-pcode-backend/src/arch_spec/hexagon/tests/arith.rs
yuv418: The output shows up in nextest when I use `RUST_LOG` appropriately, so I'm not sure what advantage using println offers here. Open to discussion though

yuv418: If you have suggestions on where to put the `println!`, I'll do it.

.github/workflows/hexagon.yml
YurBoiRene: Do they take a long time where you ONLY want to run if something definitely changes? What I am reading is that it will only run if one of the hexagon files change, not if any styx-core changes as you commented. I.e. a change in `styx/core/styx-processor/executor/mod.rs` would not run the hexagon tests in this current state.

You could do what `cargo-ci.yml` does:

```yaml
    paths-ignore:
      - "./**.md"
      - "docs/**"
```

data/test-binaries/styx-hexagon-testdata/data/build_ci.sh
```diff
@@ -0,0 +1,33 @@
...
+# system-level Hexagon support and various Hexagon peripherals. It then compiles these tests
+# using the instructions in qemu-hexagon-testing's repository. Compilation of qemu-hexagon-testing
+# requires a recent Hexagon SDK.
+#
+# The script requiers the environment variable HEXAGON_SDK_ROOT to be set. The root looks like
+# /path/to/Hexagon_SDK/VERSION. For example, /opt/Hexagon_SDK/6.5.0.0 would be a valid HEXAGON_SDK_ROOT.
+#
+# styx-hexagon-testdata uses a Dockerfile to build a Docker image with test binaries available at /testdata.
+# The test binaries are extracted from Docker and are later available for use in other styx crates.
+# During the Docker image build, this script run to build the test binaries.
```
YurBoiRene: nit: typo: change to "During the Docker image build, this script is run to build the test binaries."

.github/workflows/hexagon.yml
```diff
@@ -0,0 +1,73 @@
...
+    - name: clean
+      shell: bash
+      run: |
+        cargo clean
+
+    - name: clean
+      shell: bash
+      run: |
+        cd styx/bindings
+        cargo clean
```
YurBoiRene: I don't think this is necessary. Shouldn't be touching bindings here

YurBoiRene (changes requested):
Mostly just the `paths` filter and I think the CI will build the docker container. This is probably fine but I know we talked about making this optimization.

.github/workflows/hexagon.yml
```diff
@@ -0,0 +1,73 @@
...
+      - main
+
+concurrency:
+  # Cancels pending runs when a PR gets updated.
+  group: ${{ github.head_ref || github.run_id }}-${{ github.actor }}-${{ github.workflow }}
+  cancel-in-progress: true
+
+env:
+  HEXAGON_SDK_ROOT: "/opt/Hexagon_SDK/6.5.0.0"
+
```
YurBoiRene: Remove: not necessary anymore since it's all built in container.

YurBoiRene (approved):

PR #198, opened 2026-05-16
Nextest ci profile

YurBoiRene:
Remove comments and default values from default `nextest.toml` so it only contains relevent settings.

Added a CI profile has an extended "slow' timeout. We can add more verbose logging and other configurations here as well. The github actions should use this ci profile.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,32 +1,7 @@
-# This is the default config used by nextest. It is embedded in the binary at
-# build time. It may be used as a template for .config/nextest.toml.
-
-[store]
-# The directory under the workspace root at which nextest-related files are
-# written. Profile-specific storage is currently written to dir/<profile-name>.
-dir = "target/nextest"
-
-# This section defines the default nextest profile. Custom profiles are layered
-# on top of the default profile.
 [profile.default]
-# "retries" defines the number of times a test should be retried. If set to a
-# non-zero value, tests that succeed on a subsequent attempt will be marked as
-# non-flaky. Can be overridden through the `--retries` option.
-# Examples
-# * retries = 3
-# * retries = { backoff = "fixed", count = 2, delay = "1s" }
-# * retries = { backoff = "exponential", count = 10, delay = "1s", jitter = true, max-delay = "10s" }
 # added retries to avoid racy ipc cond var when runners
 # or dev stations are under load
 retries = 3
-
-# The number of threads to run tests with. Supported values are either an integer or
-# the string "num-cpus". Can be overridden through the `--test-threads` option.
-test-threads = "num-cpus"
-
-# The number of threads required for each test. This is generally used in overrides to
-# mark certain tests as heavier than others. However, it can also be set as a global parameter.
-threads-required = 1
 
 # Show these test statuses in the output.
 #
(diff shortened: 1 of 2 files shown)


.config/nextest.toml
```diff
@@ -41,51 +16,22 @@ threads-required = 1
...
 
 # Treat a test as leaky if after the process is shut down, standard output and standard error
 # aren't closed within this duration.
 #
 # This usually happens in case of a test that creates a child process and lets it inherit those
 # handles, but doesn't clean the child process up (especially when it fails).
-#
-# See <https://nexte.st/book/leaky-tests> for more information.
 leak-timeout = "100ms"
+global-timeout = "4m"
```
yuv418, line 44: I would document `global-timeout` and `slow-timeout` above

.config/nextest.toml
```diff
@@ -116,3 +62,9 @@ test-threads = 8
 
 [profile.quiet]
 failure-output = "never"
+
+[profile.ci]
+status-level = "skip"
+# CI runners are slower than dev machines
+slow-timeout = { period = "60s", terminate-after = 4 }
```
yuv418, line 85: I would carry over or link to documentation (if added above) here.

.config/nextest.toml
```diff
@@ -116,3 +62,9 @@ test-threads = 8
 
 [profile.quiet]
 failure-output = "never"
+
+[profile.ci]
+status-level = "skip"
```
yuv418: same here, "see status-level documentation above" or something

yuv418 (changes requested):
looks good, would re-add a few comments and gtg

.config/nextest.toml
```diff
@@ -41,52 +16,33 @@ threads-required = 1
...
 #
 # This usually happens in case of a test that creates a child process and lets it inherit those
 # handles, but doesn't clean the child process up (especially when it fails).
-#
-# See <https://nexte.st/book/leaky-tests> for more information.
 leak-timeout = "100ms"
 
+# Test runs will be terminated if they take longer than this timeout.
+# See <https://nexte.st/book/slow-tests> for more information.
+global-timeout = "4m"
```
yuv418, line 44: Should the global timeout here be the same as the slow timeout for the default profile?

.config/nextest.toml
```diff
@@ -1,33 +1,8 @@
...
-# * retries = 3
-# * retries = { backoff = "fixed", count = 2, delay = "1s" }
-# * retries = { backoff = "exponential", count = 10, delay = "1s", jitter = true, max-delay = "10s" }
 # added retries to avoid racy ipc cond var when runners
 # or dev stations are under load
 retries = 3
 
-# The number of threads to run tests with. Supported values are either an integer or
-# the string "num-cpus". Can be overridden through the `--test-threads` option.
-test-threads = "num-cpus"
```
yuv418, line 25: Will removing this make the tests run more slowly?

yuv418 (commented):
A few more comments, feel free to change stuff if you think it's necessary otherwise looks good!

.config/nextest.toml
YurBoiRene: I don't think so. global timeout refers to the whole run, the slow timeout terminate after option refers to a single test.

.config/nextest.toml
YurBoiRene: No, `num-cpus` is the [default value for this option](https://nexte.st/docs/configuration/reference/#profiletest-threads). I removed it because I thought it was overly verbose to include the default values.

PR #201, opened 2026-05-23
fix: styx-blackfin-sys build error

YurBoiRene:
On my machine (Fedora 44) this was resulting in a build error in bindgen's layout size check, causing them to be generated as opaque. I am not sure why they aren't being resolved, but we can treat them as opaque fields anyway.

I verified that his occurs on a fresh install of Fedora 44 as well. I havn't nailed down the cause but gotta be something with updated packages.


yuv418 (approved):
got the same on f44. approve

PR #202, opened 2026-05-28
fix: unicorn on fedora 44

yuv418:
I think this fixes issues with libatomic on fedora 44. But I don't know if it breaks things with other OSes/distros.


yuv418:
I'm a bit confused why it's showing a diff that shouldn't be a diff.

yuv418:
https://sourceware.org/bugzilla/show_bug.cgi?id=29455 this might be a hint as to what's going on.

YurBoiRene (changes requested):
Thanks this is good. Can you rebase on main and add a comment on the build.rs, just something like "Builds on modern distros fail with missing symbols found in libatomic. "

YurBoiRene (approved):

PR #207, opened 2026-06-18
Loop detection and shadow stack plugin

rendlur-kudu:
- Initial code for the loop detection and shadow stack plugin
- Minor changes to make the codebase compile on macOS

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ members = [
   "./examples/diy-processor",
   "./examples/fuzzer-plugin",
   "./examples/kinetis21-interrupt-angr",
+  "./examples/loop-detection-plugin",
   "./examples/multiple-processors",
   "./examples/ppc405-freertos-demo",
   "./examples/raw-processor",
@@ -380,7 +381,7 @@ opt-level = 3
 lto = "fat"
 strip = "symbols"
 debug = "full"
-split-debuginfo = "true"
+split-debuginfo = "packed"
 debug-assertions = false
 overflow-checks = false
 codegen-units = 1
--- a/styx/core/styx-cpu-pcode-backend/src/get_pcode.rs
+++ b/styx/core/styx-cpu-pcode-backend/src/get_pcode.rs
@@ -316,7 +316,17 @@ fn branch_target(class: InstructionClass, pcodes: &[Pcode]) -> Option<u64> {
 
 /// Project the Pcode ops of an instruction into an [`InstructionClass`]
 /// 
-/// This is a heuristic, so it could be improved upon
+/// This function is useful for when we need to classify the "type" of a machine instruction. For
+/// example, does this instruction correspond to a function prologue and epilogue (useful for the
+/// shadow stack plugin), or does this instruction do some comparison, etc.
+/// 
+/// However, one machine instruction lowers to several pcode ops, so we walk through all of the
+/// pcode ops that were generated and record which ones were present. We then pick a single one
+/// based on a heuristic priority (a return instruction being in the generated pcode ops indicates
+/// that the original instruction is a return instruction, even if there was a load or branch pcode op
+/// present).
+/// 
+/// This is a heuristic, so it could be improved upon.
 fn classify_pcodes(pcodes: &[Pcode]) -> InstructionClass {
     use Opcode::*;
     let mut ret = false;
--- a/styx/core/styx-cpu-pcode-backend/src/get_pcode.rs
+++ b/styx/core/styx-cpu-pcode-backend/src/get_pcode.rs
@@ -265,7 +265,7 @@ fn contains_branch_instruction(pcodes: &[Pcode]) -> bool {
 /// about detecting if an instruction is a call/return. I'm not 100% of the specifics of how
 /// Ghidra decomposes an instruction into multiple pcode ops, but there might be a better
 /// heuristic for collapsing the ops into a single classification.
-/// 
+///
 /// If the instruction cannot be decoded, it returns [`InstructionClass::Unknown`] with a length of `0`
 pub(crate) fn classify_instruction(
     cpu: &mut PcodeBackend,
@@ -297,7 +297,7 @@ pub(crate) fn classify_instruction(
 }
 
 /// Extract the statically known target of a direct call/branch
-/// 
+///
 /// Returns `None` for indirect transfers and returns
 fn branch_target(class: InstructionClass, pcodes: &[Pcode]) -> Option<u64> {
     let opcode = match class {
@@ -315,17 +315,17 @@ fn branch_target(class: InstructionClass, pcodes: &[Pcode]) -> Option<u64> {
 }
 
 /// Project the Pcode ops of an instruction into an [`InstructionClass`]
-/// 
+///
 /// This function is useful for when we need to classify the "type" of a machine instruction. For
 /// example, does this instruction correspond to a function prologue and epilogue (useful for the
 /// shadow stack plugin), or does this instruction do some comparison, etc.
-/// 
+///
 /// However, one machine instruction lowers to several pcode ops, so we walk through all of the
 /// pcode ops that were generated and record which ones were present. We then pick a single one
 /// based on a heuristic priority (a return instruction being in the generated pcode ops indicates
 /// that the original instruction is a return instruction, even if there was a load or branch pcode op
 /// present).
-/// 
+///
 /// This is a heuristic, so it could be improved upon.
 fn classify_pcodes(pcodes: &[Pcode]) -> InstructionClass {
     use Opcode::*;
@@ -373,7 +373,3 @@ fn classify_pcodes(pcodes: &[Pcode]) -> InstructionClass {
         InstructionClass::Other
     }
 }
-
-
-
-
--- a/styx/core/styx-cpu-pcode-backend/src/get_pcode.rs
+++ b/styx/core/styx-cpu-pcode-backend/src/get_pcode.rs
@@ -3,11 +3,11 @@ use log::{debug, trace};
 use smallvec::SmallVec;
 use styx_cpu_type::TargetExitReason;
 use styx_errors::{anyhow::Context, UnknownError};
-use styx_pcode::pcode::Pcode;
+use styx_pcode::pcode::{Opcode, Pcode, SpaceName};
 use styx_pcode_translator::ContextOption;
 use styx_processor::{
     core::{FetchException, HandleExceptionAction},
-    cpu::CpuBackend,
+    cpu::{CpuBackend, InstructionClass, InstructionInfo},
     event_controller::{EventController, ExceptionNumber},
     hooks::{MemFaultData, Resolution},
     memory::{MemoryOperationError, MemoryPermissions, Mmu, MmuOpError},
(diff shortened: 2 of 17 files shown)


Cargo.toml
```diff
@@ -380,7 +381,7 @@ opt-level = 3
 lto = "fat"
 strip = "symbols"
 debug = "full"
-split-debuginfo = "true"
+split-debuginfo = "packed"
```
YurBoiRene, line 384: Thanks, can you split this into a separate commit.

styx/core/styx-cpu-pcode-backend/src/get_pcode.rs
```diff
@@ -260,3 +260,110 @@ pub(crate) fn is_branching_instruction<
...
+    } else if store {
+        InstructionClass::Store
+    } else if load {
+        InstructionClass::Load
+    } else if compare {
+        InstructionClass::Compare
+    } else {
+        InstructionClass::Other
+    }
+}
```
YurBoiRene: Could you expand upon the heuristic in a comment on this function. From my reading it finds any existence of these instruction classes and then returns the one with the highest priority, starting from ret, then call, etc.

styx/core/styx-cpu-unicorn-backend/build.rs
```diff
@@ -3,5 +3,12 @@
 //! This file exists to add libatomic when linking.
 
 fn main() {
-    println!("cargo:rustc-link-lib=atomic");
+    // MacOS does not need libatomic, will fail to compile if included on MacOS
+    if std::env::var("CARGO_CFG_TARGET_OS").unwrap() == "linux" {
+        println!("cargo:rustc-link-lib=atomic");
+    }
```
YurBoiRene, line 9: LOL smh my head this libatomic stuff is annoying.

Can you split into another commit?

styx/plugins/styx-debug-tools/src/loops.rs
```diff
@@ -0,0 +1,224 @@
...
+    fn call(&mut self, mut proc: CoreHandle, _address: u64, _size: u32) -> Result<(), UnknownError> {
+        let report = self.handle.read(|stack| self.counters.observe_stack(stack));
+        if let Some(report) = report {
+            self.report_loop(&report, &mut proc);
+        }
+        Ok(())
+    }
+}
+
+#[derive(Default, serde::Deserialize)]
```
YurBoiRene: What's the point of the serde impls?

styx/core/styx-cpu-pcode-backend/src/get_pcode.rs
```diff
@@ -260,3 +260,110 @@ pub(crate) fn is_branching_instruction<
...
+    mmu: &mut Mmu,
+    ev: &mut EventController,
+) -> InstructionInfo {
+    let mut pcodes = Vec::with_capacity(16);
+
+    let mut helper = cpu.pcode_generator.helper.take().unwrap();
+    let ctx_opts = helper.pre_fetch(cpu).unwrap();
+    cpu.pcode_generator.helper = Some(helper);
+
+    match get_pcode_at_address(cpu, addr, &mut pcodes, &ctx_opts, mmu, ev) {
```
YurBoiRene, line 282: Yeah this will be super slow. Pcode translation is about 50% of the execution time in the pcode backend (the other 50% being actual instruction execution). That's okay and should improve when we get a caching layer in there.

I need to take a closer look, I am afraid this may call hooks (mem read, invalid instruction, memory fault) which we wouldn't want to do here.

styx/core/styx-processor/src/cpu/backend.rs
```diff
@@ -152,4 +265,20 @@ pub trait CpuBackend: Debug + Hookable + Send {
...
+        event_controller: &mut EventController,
+    ) -> InstructionInfo {
+        let _ = (addr, mmu, event_controller);
+        InstructionInfo {
+            class: InstructionClass::Unknown,
+            length: 0,
+            target: None,
+        }
+    }
 }
```
YurBoiRene, line 283: Maybe this should return an `Option<InstructionInfo>` so `None` can be the unimplemented state.

YurBoiRene (changes requested):
Thank you! I like this changset. 

Just split up the MacOS build changes into separate commits and there's some CI faliures.

I want to mull over the addition to the CpuBackend.

Cargo.toml
rendlur-kudu: done!

styx/core/styx-cpu-unicorn-backend/build.rs
rendlur-kudu: done!

styx/plugins/styx-debug-tools/src/loops.rs
rendlur-kudu: Needed for the `styx_uconf::register_component_config!()` macro

styx/core/styx-processor/src/cpu/backend.rs
rendlur-kudu: Yea, that's a better idea. Changed!

styx/core/styx-cpu-pcode-backend/src/get_pcode.rs
rendlur-kudu: done!

styx/plugins/styx-debug-tools/src/loops.rs
```diff
@@ -0,0 +1,235 @@
...
+    threshold: u64,
+    /// If true, stop emulation when loop is reported
+    #[serde(default)]
+    halt_on_report: bool,
+    /// Invoked when loop is reported
+    #[serde(skip)]
+    on_report: Option<LoopCallback>,
+    #[serde(skip)]
+    shadow_stack: Option<ShadowStackHandle>,
+}
```
YurBoiRene, line 204: Okay this is a bit nit, but I think it would make for a better interface for users. Feel free to push back.

Instead of directing them to use `new` with several arguments, I'm thinking instruct them to use `default` with a builder pattern.

I.e.
```rust
LoopDetectionPlugin::default().with_threshold(5).with_halt_on_report(true)
```
and then have reasonable defaults. That way the user can just do `LoopDetectionPlugin::default()` in the typical case.

I like "halt_on_report". I foresee the most typical use case of this plugin to be to log a message at the info level. The behavior of the plugin after detection could be one enum, instead of the separate hook and halt hook:

With this I would make the default be `DetectionBehavior::Log(log::Level::Info)`
```rust
pub enum DetectionBehavior {
    /// Just does cpu.stop
    Halt,
    Custom(LoopCallback),
    ///  log loop to logger
    Log(log::Level),
    /// Throw an error in the hook
    Error,
}
```

The `DetectionBehavior` is less of an improvement than the builder api, but up to you. Let me know your thoughts.

styx/plugins/styx-debug-tools/src/loops.rs
```diff
@@ -0,0 +1,235 @@
...
+        Self {
+            head_addr,
+            frame_addr,
+            iters,
+            stack_depth,
+        }
+    }
+}
+
+pub type LoopCallback = Box<dyn FnMut(&LoopReport) + Send>;
```
YurBoiRene: I would add CoreHandle to this callback.

styx/plugins/styx-debug-tools/src/loops.rs
```diff
@@ -0,0 +1,235 @@
...
+    on_report: Option<LoopCallback>,
+    #[serde(skip)]
+    shadow_stack: Option<ShadowStackHandle>,
+}
+
+styx_uconf::register_component_config!(register plugin: id = loops, component = LoopDetectionPlugin);
+
+impl LoopDetectionPlugin {
+    pub fn new(
+        threshold: u64,
```
YurBoiRene: I might make `threashold` a `NonZeroU64` since it can't be zero (noted by one of your comments below). At the very least document the fact that it has a minimum of 1 and will be overwritten

styx/plugins/styx-debug-tools/src/loops.rs
```diff
@@ -0,0 +1,235 @@
...
+        halt_on_detection: bool,
+        on_detection: Option<LoopCallback>,
+        shadow_stack: Option<ShadowStackHandle>,
```
YurBoiRene: Does the user ever have to provide this, or does the plugin make it in all cases. Probably document this for the plugin

styx/plugins/styx-debug-tools/src/shadow_stack.rs
```diff
@@ -0,0 +1,324 @@
...
+            let term =
+                classify_block_terminator(core.cpu, address, size, core.mmu, core.event_controller);
+            if (term.is_none() || term.unwrap().flow == ControlFlowType::Unknown)
+                && !warned_weird_state
+            {
+                warn!(
+                    "ShadowStack: backend cannot classify control flow; \
+                    the shadow stack might get weird man!"
+                );
+                warned_weird_state = true;
```
YurBoiRene, line 280: Would this work at all? If not it's okay to fatal error imo

styx/plugins/styx-debug-tools/src/loops.rs
```diff
@@ -0,0 +1,235 @@
...
+    ) -> Result<Box<dyn Plugin>, UnknownError> {
+        let handle = match self.shadow_stack.take() {
+            Some(handle) => {
+                if !handle.has_updater() {
+                    return Err(anyhow!(
+                        "LoopDetectionPlugin: the provided shadow stack does not have an updater.\
+                        Add the ShadowStackPlugin before the LoopDetectionPlugin"
+                    ));
+                }
+                handle
```
YurBoiRene, line 272: Styx needs a dependent plugin api :) will make an issue for that

YurBoiRene (changes requested):
Nice :)

styx/plugins/styx-debug-tools/src/loops.rs
rendlur-kudu: I did not know that existed, that's a much better idea!

styx/plugins/styx-debug-tools/src/loops.rs
rendlur-kudu: I fully agree with the first point, and I checked to see if we should use that pattern anywhere else for these two plugins (not really since the only other one the user would have access to would be the ShadowStackHandle, which doesn't have any fields, and ShadowStackPlugin, which needs to be fed a handle).

For the second one, what If we want to do multiple things? Like we want to Halt, do a custom callback, and log it? Should the DetectionBehavior be a bitmap, or do you think that it's not reasonable that a user would want any two being active.

styx/plugins/styx-debug-tools/src/loops.rs
rendlur-kudu: done!

styx/plugins/styx-debug-tools/src/loops.rs
rendlur-kudu: this is fixed since we now use the builder pattern

styx/plugins/styx-debug-tools/src/shadow_stack.rs
rendlur-kudu: I changed it so it would panic, but I also kept the previous code as a comment.

I was initially thinking of making the code error out, but I chose warning the user because I'd imagine that the loop stuff isn't a top priority, and it would be very annoying for the plugin to keep producing a panic. AFAIK, there currently isn't a way to just shut off a plugin, perhaps this could be a feature?

styx/plugins/styx-debug-tools/src/loops.rs
rendlur-kudu: ack

YurBoiRene: #209

styx/plugins/styx-debug-tools/src/shadow_stack.rs
```diff
@@ -249,17 +249,26 @@ impl ShadowStackHandle {
...
-    let mut warned_weird_state = false;
+    // let mut warned_weird_state = false;
     proc.core
         .cpu
         .add_hook(StyxHook::block(move |core: CoreHandle, address, size| {
             let term =
                 classify_block_terminator(core.cpu, address, size, core.mmu, core.event_controller);
+
+            if term.is_none() || term.unwrap().flow == ControlFlowType::Unknown {
+                panic!("ShadowStack: backend cannot classify control flow");
```
YurBoiRene: Instead of panic, can you make this `return Err(anyhow!("error text"))`. It will end up panicing anyway but this should be nicer for users who use styx as a library.

styx/plugins/styx-debug-tools/src/shadow_stack.rs
YurBoiRene: Ah makes sense. I think the panic with the warning commented out is good for now, if someone has a use case where they expect it to error then they can maybe add an option to the plugin to panic vs just warn.

btw I added a comment to use `return Err(anyhow!("error txt"))` instead of `panic!()` that way library users don't *need* to crash, they can handle the error as they see fit when it propagates to the `emulate()` call.

styx/plugins/styx-debug-tools/src/loops.rs
YurBoiRene: Good point. It doesn't seem to hard to make the detection_behavior be a Vec of these. Just be mindful of the api that this creates for the builder. I assume the methods will be like `add_behavior()`, `clear_behaviors()`, or maybe like common ones like `with_halt_and_log()`, `with_log()` (clears and sets the only behavior to the one specified)

styx/plugins/styx-debug-tools/src/shadow_stack.rs
rendlur-kudu: ack

styx/plugins/styx-debug-tools/src/shadow_stack.rs
rendlur-kudu: done!

maybe we could add add a method for the plugin traits that lets the user run a function in case of an error

styx/plugins/styx-debug-tools/src/loops.rs
rendlur-kudu: done!

YurBoiRene:
merged in #212

PR #211, opened 2026-07-14
perf: parallelize styx-sla build

YurBoiRene:
This spawns separate worker processes to compile each spec. Multi-processing the compilation of the slas significantly speeds up compile time for this crate. I'm seeing about 4x improvement. This is needed because libsleigh is non-reentrant.

Assisted-by: Claude:claude-opus-4-8

### DIFF
--- a/styx/core/styx-cpu-pcode-backend/styx-sla/build.rs
+++ b/styx/core/styx-cpu-pcode-backend/styx-sla/build.rs
@@ -1,10 +1,35 @@
 // SPDX-License-Identifier: BSD-2-Clause
+
+//! Build script for `styx-sla`: turns the checked-in SLEIGH processor
+//! definitions into compiled `.sla` artifacts plus generated Rust bindings.
+//!
+//! Pipeline (see [`main`]):
+//! 1. Copy the `processors/` tree (Ghidra + custom slaspecs) into `OUT_DIR`,
+//!    then apply the unified-diff patches from `patches/` onto that copy.
+//! 2. For every arch feature enabled at build time (see [`SPECS`] and the
+//!    `arch_*` cargo features), compile each `.slaspec` into a `.sla` with
+//!    Ghidra's SLEIGH compiler ([`process`]).
+//! 3. Load each `.sla` to extract its user-ops and emit Rust into
+//!    `$OUT_DIR/sla_artifacts.rs`: a `SlaSpec`/`SlaUserOps` impl per spec plus a
+//!    `<Spec>UserOps` enum. `src/lib.rs` `include!`s that file.
+//!
+//! ## Concurrency
+//! Compiling a slaspec goes through Ghidra's C++ SLEIGH compiler, which
+//! relies on process-global lexer/parser state and therefore cannot be run
+//! concurrently within a single process (threads would corrupt each other).
+//! To use more than one core we instead spawn one *child process* per
+//! spec by re-executing this same build-script binary in "worker" mode. Each
+//! worker compiles exactly one spec and writes the generated rust snippet
+//! to a file which the parent then concatenates them.
+//!
 use heck::ToUpperCamelCase;
 use std::{
+    collections::VecDeque,
     env,
-    fs::File,
     io::{BufWriter, Write},
     path::{Path, PathBuf},
+    sync::{Arc, Mutex},
+    thread::JoinHandle,
 };
 use styx_pcode_sleigh_backend::Sleigh;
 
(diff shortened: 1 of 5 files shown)


styx/core/styx-cpu-pcode-backend/styx-sla/build.rs
```diff
@@ -1,8 +1,30 @@
...
+//!    Ghidra's SLEIGH compiler ([`process`]).
+//! 3. Load each `.sla` to extract its user-ops and emit Rust into
+//!    `$OUT_DIR/sla_artifacts.rs`: a `SlaSpec`/`SlaUserOps` impl per spec plus a
+//!    `<Spec>UserOps` enum. `src/lib.rs` `include!`s that file.
+//!
+//! ## Concurrency
+//! Compiling a slaspec goes through Ghidra's C++ SLEIGH compiler, which
+//! relies on process-global lexer/parser state and therefore cannot be run
+//! concurrently within a single process (threads would corrupt each other).
+//! To use more than one core we instead fan out one *child process* per
```
yuv418: nit: I don't like the word "fan out," as it feels like it takes a little more mental process to figure out what it means in this context. I think "spawn" would work better.

styx/core/styx-cpu-pcode-backend/styx-sla/build.rs
```diff
@@ -188,12 +210,22 @@ fn process(spec_path: &str, generated_code: &mut BufWriter<File>) {
...
 }
 
+/// Env var used to re-invoke this build script binary in "worker" mode. When
+/// set, its value is the single slaspec path to compile (see [`worker_main`]).
+const WORKER_SPEC_ENV: &str = "STYX_SLA_WORKER_SPEC";
+/// Env var giving the worker the absolute path it should write generated rust to.
+const WORKER_OUT_ENV: &str = "STYX_SLA_WORKER_OUT";
+
 fn main() {
+    // Check if this process is a worked (see file level documentation).
```
yuv418: typo? worker not worked

styx/core/styx-cpu-pcode-backend/styx-sla/build.rs
```diff
@@ -204,31 +236,113 @@ fn main() {
...
 
     println!("cargo::rerun-if-changed=processors/");
     println!("cargo::rerun-if-changed=patches/");
     println!("cargo::rerun-if-changed=src/lib.rs");
 }
 
+/// Worker entry point: compile a single spec and write the generated rust to the
+/// path given in [`WORKER_OUT_ENV`]. Runs in a dedicated child process so that
+/// the non-reentrant C++ SLEIGH compiler can run in parallel across specs.
+fn worker_main(spec: &str) {
```
yuv418: I might be misunderstanding, but why have `process` return a String if its only invocation ultimately writes to a file anyway?  It feels like the entire functionality can go into `process` then. Maybe a nitpick in structure though.

styx/core/styx-cpu-pcode-backend/styx-sla/build.rs
```diff
@@ -204,31 +236,113 @@ fn main() {
...
+    for (snippet_path, spec) in snippet_paths.iter().zip(enabled_specs.iter()) {
+        let snippet = std::fs::read(snippet_path)
+            .unwrap_or_else(|e| panic!("missing worker output for `{spec}`: {e}"));
+        generated_code.write_all(&snippet).unwrap();
+    }
     generated_code.flush().unwrap();
 
     println!("cargo::rerun-if-changed=processors/");
     println!("cargo::rerun-if-changed=patches/");
     println!("cargo::rerun-if-changed=src/lib.rs");
```
yuv418: Random thought, but does the slaspec need to be regenerated in this case anyway?

yuv418 (changes requested):
Neat idea, I like it! Just a few comments. This will save a lot of dev time 😁

styx/core/styx-cpu-pcode-backend/styx-sla/build.rs
YurBoiRene: Valid, done

styx/core/styx-cpu-pcode-backend/styx-sla/build.rs
YurBoiRene: Nice. I refactored into a "JobInputs" that has the two env variables. Then process just takes those inputs and does everything, including write to the final output file. This removes the worker_main too

styx/core/styx-cpu-pcode-backend/styx-sla/build.rs
YurBoiRene: Ah probably not if lib.rs changes, yes. Nice catch.

YurBoiRene:
I had to make some major changes. According to the jobserver protocol, the running "make thread" (in this case the styx-sla build script) implicitly has a token, but in the previous iteration we weren't using this token. In the case of `-j 1` this means we won't make any progress as we will never acquire any more token.

This changes it so that the main build script runs through the jobs sequentially while a separate thread tries to acquire and dispatch worker processes via job tokens. Unfortunately this makes the implementation more complex but it is necessary. Still the same performance characteristics, and theoretically should be slightly better since we are using the build script job token now.

styx/core/styx-cpu-pcode-backend/styx-sla/build.rs
```diff
@@ -204,29 +307,142 @@ fn main() {
...
+    let self_exe = Arc::new(self_exe);
+
+    // A build script is itself a "make thread" that already holds one implicit
+    // jobserver token, so it may always run *one* unit of work without asking
+    // the jobserver for anything.
+    //
+    // Instead we split the work two ways:
+    //   * The helper thread requests extra tokens from the jobserver; each
+    //     granted token dispatches one job onto its own worker thread.
+    //   * The main thread, using our implicit token, drains the same queue
```
yuv418, line 363: So does this means that in the single threaded case, the helper threads won't run anything (and the build script will do all the work)?

styx/core/styx-cpu-pcode-backend/styx-sla/build.rs
```diff
@@ -204,29 +307,142 @@ fn main() {
...
+        helper.request_token();
+    }
+
+    // Drain the queue on the main thread using our implicit token.
+    loop {
+        // Note for future devs, this looks like it could be rewritten as a
+        // `while let Some(job) = jobs.lock().unwrap().pop_front() {}`,
+        // however, this will hold the lock while inide the loop body,
+        // which is not what we want here.
+        let job = jobs.lock().unwrap().pop_front();
```
yuv418, line 408: nit: I am pretty sure that the unwrap()ped value is dropped after pop_front, but ifdoing an explicit drop or using a scope to get the job might make it more obvious that the lock is not held throughout the match statement following. But very minor

yuv418 (approved):
Looks good, just a few comments that are minor. feel free to change or just merge atp. Exciting change 👀

styx/core/styx-cpu-pcode-backend/styx-sla/build.rs
YurBoiRene: Correct, they will sit blocked, waiting for a token that never comes.

